PR #58, opened 2023-09-25
Grid number of GPTQ matrix reconstruction

chu-tianxiang:
If I'm not mistaken, each `reconstruct_gptq_kernel` thread reconstructs 4-columns, so it suffices to use fewer blocks.


turboderp:
Yes, this was an oversight since the EXL2 kernel hasn't received the int4 optimization yet.

Issue #56, opened 2023-09-23
publish the releases to github


happysalada:
can you publish the releases to github ?

Nikita-Sherstnev:
Publishing package to pip or conda with precompiled extension would be also nice.

turboderp:
There is a pip package already (`pip install exllamav2` works up to the latest commit with a version number). It looks like @jllllll already has a fork that allows permanently installing the extension, though there are no pre-built binaries yet. At some point I do plan to petition them to bring it all into the main repo, because especially the way they did it for V2 where the JIT compilation is still supported would be really nice to just have. I think the holdup is I just haven't familiarized myself with the workflows and whatever else is required to generate all the pre-built wheels with every release.

jllllll:
I can open a PR that includes workflows, though I have no experience with pushing wheels to pypi automatically with them.

turboderp:
That would be awesome. I doubt pushing wheels to PyPI could be much more involved than using twine.

jllllll:
Doing some testing and I'm not sure how to go about uploading the pre-compiled wheels to PyPI without requiring that users specify the exact version to install: `pip install exllamav2==0.0.3+cu118`

pip will view any version tags like that as being a newer version and will consequently prefer to install that wheel unless explicitly told what wheel version to install.

PR #61, opened 2023-09-25
Build wheels with pre-compiled CUDA kernels

jllllll:
Resolves: https://github.com/turboderp/exllamav2/issues/56

---
This makes JIT compiling optional and can be opted into by setting the `EXLLAMA_NOCOMPILE` environment variable on install.

This also includes 2 workflows:
- `build-wheels.yml`
  - This workflow will build wheels and sdist and upload them as workflow artifacts.
  - Useful for testing as you don't need to worry about overwriting working wheels with broken ones.
  - Calls `build-wheels-rocm.yml` to build ROCm wheels at the end.
- `build-wheels-release.yml`
  - This workflow will do the same as the first and will also upload them to GitHub releases that it will create if needed.
  - Parses `setup.py` to determine package version that it will then use as the release tag to upload to.
  - Currently configured to overwrite wheels already uploaded in a release.
  - Also uploads to PyPI. Current code for this is based on the info here:
  https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
  This may need to be adjusted as I have no experience working with PyPI and typically just make my own package indices.
  Regardless, an option is provided when running the workflow to not upload to PyPI.
  - Calls `build-wheels-release-rocm.yml` to build ROCm wheels just before building the sdist and JIT compile wheel.
- `build-wheels-rocm.yml`
  - This workflow will build wheels using ROCm 5.6. This version can be changed by editing the variable in the SDK installation step.
- `build-wheels-release-rocm.yml`
  - The same as the previous but also includes the upload functionality of `build-wheels-release.yml`.

### DIFF
--- a/.github/workflows/build-wheels-release-rocm.yml
+++ b/.github/workflows/build-wheels-release-rocm.yml
@@ -1,0 +1,109 @@
+name: Build ROCm 5.4.2 Wheels & Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      pypi:
+        description: 'Upload wheels to PyPI? 1 = yes, 0 = no'
+        default: '1'
+        required: true
+        type: string
+  workflow_call:
+    inputs:
+      pypi:
+        description: 'Upload wheels to PyPI? 1 = yes, 0 = no'
+        default: '0'
+        required: true
+        type: string
+
+permissions:
+  contents: write
+
+jobs:
+  build_wheels:
+    name: Build ROCm wheel for Python ${{ matrix.pyver }}
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        pyver: ["3.8", "3.9", "3.10", "3.11"]
+
+    steps:
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@v1.2.0
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: false
+          swap-storage: false
+    
+      - uses: actions/checkout@v3
+          
+      - name: Install ROCm SDK
+        run: |
+          export ROCM_VERSION=5.6
+          
+          [ ! -d /etc/apt/keyrings ] && sudo mkdir --parents --mode=0755 /etc/apt/keyrings
+          wget https://repo.radeon.com/rocm/rocm.gpg.key -O - | gpg --dearmor | sudo tee /etc/apt/keyrings/rocm.gpg > /dev/null
+          echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/rocm.gpg] https://repo.radeon.com/rocm/apt/$ROCM_VERSION focal main" | sudo tee --append /etc/apt/sources.list.d/rocm.list
+          echo -e 'Package: *\nPin: release o=repo.radeon.com\nPin-Priority: 600' | sudo tee /etc/apt/preferences.d/rocm-pin-600
+          sudo apt update
+          sudo apt install rocm-hip-sdk -y
+          echo "/opt/rocm/bin" >> $GITHUB_PATH
+          echo "ROCM_PATH=/opt/rocm" >> $GITHUB_ENV
+          echo "ROCM_VERSION=$ROCM_VERSION" >> $GITHUB_ENV
+          echo "USE_ROCM=1" >> $GITHUB_ENV
+          
+      - uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.pyver }}
+          
+      - name: Install Dependencies
+        run: |
+          pip3 install torch\<=2.1.0 --pre --index-url="https://download.pytorch.org/whl/rocm$ROCM_VERSION" --extra-index-url="https://download.pytorch.org/whl/nightly/rocm$ROCM_VERSION"
+          pip3 install build wheel safetensors sentencepiece ninja
+        
+      - name: Build Wheel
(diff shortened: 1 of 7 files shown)


jllllll:
I'm not sure how to go about uploading the pre-compiled wheels to PyPI without requiring that users specify the exact version to install: `pip install exllamav2==0.0.3+cu118`

pip will view any version tags like that as being a newer version and will consequently prefer to install that wheel unless explicitly told what wheel version to install.

With these wheels available in the same place:
- `exllamav2-0.0.3-py3-none-any.whl`
- `exllamav2-0.0.3+cu117-cp310-cp310-win_amd64.whl`
- `exllamav2-0.0.3+cu118-cp310-cp310-win_amd64.whl`

This will install the JIT compile version: `pip install exllamav2===0.0.3`
This will install the pre-compiled CUDA 11.7 version: `pip install exllamav2==0.0.3+cu117`

All of these will install the pre-compiled CUDA 11.8 version:
- `pip install exllamav2==0.0.3+cu118`
- `pip install exllamav2==0.0.3`
- `pip install exllamav2`

jllllll:
Currently still testing building ROCm wheels and can I add them to this PR when done.

turboderp:
I've done a bit of research, by which I mean I asked ChatGPT, and it seems convinced that `pip` simply ignores local version identifiers by default. So unless explicitly stated, "0.0.3", "0.0.3+cu117" and "0.0.3+cu118" would all match "==0.0.3", and `pip` will install whatever version has a more recent timestamp.

If it's not hallucinating, a fix could be to simply make sure the "0.0.3" is always uploaded last. If that's doable.

jllllll:
Should be possible. Also, having some issues getting exllamav2 to build with ROCm 5.4.2. Is there a minimum ROCm version that this is compatible with?

jllllll:
Local testing shows that the file time doesn't change what pip selects, but this may only apply to `--find-links`.

turboderp:
That sounds great, yeah.

jllllll:
Added ROCm workflows that the main 2 workflows will execute. They can also be used on their own if needed.

If they cause issues with the build process in the future, then you may want to remove the workflow call from the main workflow.

turboderp:
I'll try it out.

turboderp:
The build actions seem to work great. Thank you for the contribution.

Sadder news is that it looks like an uphill battle with PyPI. For one thing, linux_x86_64 is an unsupported platform tag. According to `auditwheel` the "manylinux_2_27_x86_64" tag would work instead, so maybe it could resolved that way. But  the local version identifiers aren't supported either.

jllllll:
That's unfortunate. The `manylinux` tag could be done without too much issue, but the lack of local version identifier support greatly complicates things.

turboderp:
Yeah, it's a shame. It looks like it was supported at some point but not anymore? Idk, pip versioning is a whole thing apparently.

I guess the next best thing would be dividing the wheels up into separate directories. One for cu117, one for cu118 and so on. Like Torch does (https://download.pytorch.org/whl/cu118 etc.) I'll experiment later to see if github can accommodate something like that for releases. Not sure how much of a hack [this](https://github.com/astariul/github-hosted-pypi) is, but it looks interesting.

At any rate, at least the PyPI package that I can upload will now build and install `exllamav2-ext` by default, which seems to be preferred, and there are prebuilt wheels in the releases. Which is still progress.

jllllll:
A GitHub-hosted package index is exactly what I do in this repo and several others:
https://github.com/jllllll/llama-cpp-python-cuBLAS-wheels
https://jllllll.github.io/llama-cpp-python-cuBLAS-wheels

It is needed in that repo in particular due to the distribution of literally thousands of wheels.

zhangyuhanjc:
> I'm not sure how to go about uploading the pre-compiled wheels to PyPI without requiring that users specify the exact version to install: `pip install exllamav2==0.0.3+cu118`
> 
> pip will view any version tags like that as being a newer version and will consequently prefer to install that wheel unless explicitly told what wheel version to install.
> 
> With these wheels available in the same place:
> 
> * `exllamav2-0.0.3-py3-none-any.whl`
> * `exllamav2-0.0.3+cu117-cp310-cp310-win_amd64.whl`
> * `exllamav2-0.0.3+cu118-cp310-cp310-win_amd64.whl`
> 
> This will install the JIT compile version: `pip install exllamav2===0.0.3` This will install the pre-compiled CUDA 11.7 version: `pip install exllamav2==0.0.3+cu117`
> 
> All of these will install the pre-compiled CUDA 11.8 version:
> 
> * `pip install exllamav2==0.0.3+cu118`
> * `pip install exllamav2==0.0.3`
> * `pip install exllamav2`

is there any pre-compiled version for linux_x86_64 ?

zhangyuhanjc:
Some additional info when i compile exllama v2

these are the env i can successfully run 
 Driver Version: 530.41.03    CUDA Version: 12.1 
 Driver Version: 535.54.03    CUDA Version: 12.2
these are the env i fail to run 
Driver Version: 510.47.03    CUDA Version: 11.6 ([CUDA initialization: Unexpected error from cudaGetDeviceCount())
Driver Version: 510.47.03    CUDA Version: 12.1 (identifier "__hmul" is undefined]

turboderp:
> is there any pre-compiled version for linux_x86_64 ?

Yes. The prebuilt wheels are currently available here: https://github.com/turboderp/exllamav2/releases/

PR #71, opened 2023-09-29
Code highlighting in chat CLI

SinanAkkoyun:
Colored code in the CLI chat!

image

This PR focuses around the implementation of real-time streaming codeblocks with the `examples/chatcode.py` CLI chat script. Via pyterm, I stream the codeblocks as floating boxes with black background, similar to the ChatGPT interface.

This makes debugging code models much more pleasing :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,17 @@ pip install -r requirements.txt
 python test_inference.py -m <path_to_model> -p "Once upon a time,"
 ```
 
-For now, a simple console chatbot is included. Run it with:
+A simple console chatbot is included. Run it with:
 
 ```
 python examples/chat.py -m <path_to_model> -mode llama
 ```
+
+For a chat with colored code, run:
+```
+python examples/chatcode.py -m <path_to_model> -mode llama
+```
+
 
 The `-mode` argument chooses the prompt format to use. `llama` is for the Llama(2)-chat finetunes, while `codellama`
 probably works better for CodeLlama-instruct. `raw` will produce a simple chatlog-style chat that works with base 
--- a/examples/chatcode.py
+++ b/examples/chatcode.py
@@ -279,6 +279,8 @@ def print_code_block(chunk):
     # Update the code block text
     code_block_text += chunk
 
+    code_block_text = re.sub(r'```.*?\n', '', code_block_text, flags=re.DOTALL)
+
     # Split updated text into lines and find the longest line
     lines = code_block_text.split('\n')
     max_length = max(len(line) for line in lines)
--- a/examples/chatcode.py
+++ b/examples/chatcode.py
@@ -1,0 +1,398 @@
+import os
+import re
+import sys
+from io import StringIO
+
+from pygments import highlight
+from pygments.formatter import Formatter
+from pygments.formatters.terminal import TerminalFormatter
+from pygments.lexers import get_lexer_by_name, guess_lexer
+from pygments.style import Style
+from pygments.styles.default import DefaultStyle
+from pygments.token import Token
+from pygments.util import ClassNotFound
+
+import shutil
+
+# Append the parent directory to sys.path
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+
+from exllamav2 import(
+    ExLlamaV2,
+    ExLlamaV2Config,
+    ExLlamaV2Cache,
+    ExLlamaV2Tokenizer,
+    model_init,
+)
+
+import argparse
+import torch
+
+from exllamav2.generator import (
+    ExLlamaV2StreamingGenerator,
+    ExLlamaV2Sampler
+)
+
+# Options
+
+parser = argparse.ArgumentParser(description = "Simple Llama2 chat example for ExLlamaV2")
+parser.add_argument("-mode", "--mode", choices = ["llama", "raw", "codellama"], help = "Chat mode. Use llama for Llama 1/2 chat finetunes.")
+parser.add_argument("-un", "--username", type = str, default = "User", help = "Username when using raw chat mode")
+parser.add_argument("-bn", "--botname", type = str, default = "Chatbort", help = "Bot name when using raw chat mode")
+parser.add_argument("-sp", "--system_prompt", type = str, help = "Use custom system prompt")
+
+parser.add_argument("-temp", "--temperature", type = float, default = 0.95, help = "Sampler temperature, default = 0.95 (1 to disable)")
+parser.add_argument("-topk", "--top_k", type = int, default = 50, help = "Sampler top-K, default = 50 (0 to disable)")
+parser.add_argument("-topp", "--top_p", type = float, default = 0.8, help = "Sampler top-P, default = 0.8 (0 to disable)")
+parser.add_argument("-typical", "--typical", type = float, default = 0.0, help = "Sampler typical threshold, default = 0.0 (0 to disable)")
+parser.add_argument("-repp", "--repetition_penalty", type = float, default = 1.1, help = "Sampler repetition penalty, default = 1.1 (1 to disable)")
+parser.add_argument("-maxr", "--max_response_tokens", type = int, default = 1000, help = "Max tokens per response, default = 1000")
+parser.add_argument("-resc", "--response_chunk", type = int, default = 250, help = "Space to reserve in context for reply, default = 250")
+
+# Initialize model and tokenizer
+
+model_init.add_args(parser)
+args = parser.parse_args()
+model_init.check_args(args)
+model_init.print_options(args)
+model, tokenizer = model_init.init(args)
+
+# Create cache
+
+cache = ExLlamaV2Cache(model)
+
+# Prompt templates
+
+username = args.username
+botname = args.botname
+system_prompt = args.system_prompt
+mode = args.mode
+
+if mode == "llama" or mode == "codellama":
+
+    if not system_prompt:
+
+        if mode == "llama":
+
+            system_prompt = \
+            """You are a helpful, respectful and honest assistant. Always answer as helpfully as possible, while being safe.  """ + \
+            """Your answers should not include any harmful, unethical, racist, sexist, toxic, dangerous, or illegal content. """ + \
+            """Please ensure that your responses are socially unbiased and positive in nature."""
+
+        elif mode == "codellama":
+
+            system_prompt = \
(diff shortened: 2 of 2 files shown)


SinanAkkoyun:
If you want, I can copy this into `chat.py`, but I didn't want to bloat it without your approval

turboderp:
This is very nice. I actually had it on my todo list to find some sort of syntax highlighting library to use for this exactly. It's not perfect, I think:

Screenshot_20230929_232601

But that's something to fix down the line I guess. I know even ChatGPT struggles with formatting code as it's streaming in. Hell, I've seen Visual Studio mess up syntax highlighting sometimes, I guess it's a nontrivial problem.

In any case, I'll include it as is, then probably merge it into the main chat example later as a (default) option.

Thanks a bunch :)

SinanAkkoyun:
Wohooo, I am super glad you like it! :) 🎉❤️

And thank you for the whole repo, your nice support and your dedication to enhance LLM inference for everyone!

SinanAkkoyun:
@turboderp Sorry, I first assumed the screenshot was your attempt, that's why I respond so late. Regarding your issue; what was your exact prompt? I have tested it across multiple code lengths and languages and could not replicate your result.

SinanAkkoyun:
Regarding your screenshot, is your default color scheme red? When I try C# code it looks like this, white:
image

Not perfect by any means (Java also doesn't look too pretty), but that's up to pygments not having good syntax highlighters for these languages. If anyone finds add-ons for pygment I would be happy to upgrade the highlighters :)

turboderp:
Nope, most code comes out white. And that block was morphing a bit while updating, so maybe it was just too long?

SinanAkkoyun:
Hmm that's odd, I will OCR your code and look if it's a pygment issue when I find time
(but it theoretically must, because my code just processes everything around the codeblock for pretty formatting and injects the black background, maybe I did an oopsie with the ANSI background injection)

turboderp:
I mangled it a bit and wrapped it in a class in `examples/chat_formatting.py`, so I can use it from the main chat script without cluttering it too much. I'll do some more refactoring at some point I guess, to build out with more formatters for lists, BBCode, what have you. Eventually it'll probably morph into some sort of unified backend for console and web UI chats.

I think the lexer autodetection is broken, though. It seems kind of random what syntax highlighting it applies, and I can't get it to format comments in any language.

Screenshot_20231001_124623

It also looks like calling `guess_lexer` on every token is a little slow, so maybe it makes more sense to call it at the end of every line instead?

But anyway this is fine as something to build on.

turboderp:
(Turns out the black background was missing because of my Konsole profile, but now the output looks the same in gnome-terminal as well, so I think I have Konsole dialed in at least.)

SinanAkkoyun:
> I mangled it a bit and wrapped it in a class in examples/chat_formatting.py, so I can use it from the main chat script without cluttering it too much. I'll do some more refactoring at some point I guess, to build out with more formatters for lists, BBCode, what have you. Eventually it'll probably morph into some sort of unified backend for console and web UI chats.

How awesome! Tell me if I can help you out or do some reptitive work for you


> I think the lexer autodetection is broken, though. It seems kind of random what syntax highlighting it applies, and I can't get it to format comments in any language.

I see, at the moment I don't have much time, but given your assessment I'd like to implement the following:
- Sometimes the model specifies the language, use that (and/or incorporate a small prompt to ensure lang specification)
- Detect comments-only and lists etc by hand (always treat // only codeblocks as C++ or something)
(- Maybe improve on Java and C# formatter)

When I find time in a couple of days I will implement that

turboderp:
Well, I have some ideas, and I was going to look into this anyway, so it's fine. I did some quick digging and it looks like `pygments` is a little crude when guessing what lexer to use. Here is how it detects C++, for instance:

```python
    def analyse_text(text):
        if re.search('#include <[a-z_]+>', text):
            return 0.2
        if re.search('using namespace ', text):
            return 0.4
```

So that won't really work for code snippets. Also it supports a *lot* of obscure formats so maybe that could be narrowed down a bit for more accurate results. I'm looking into it.

SinanAkkoyun:
> So that won't really work for code snippets.

Ah I see, that's really low effort lang guessing lol
I also found `guesslang`, but it needs tensorflow installed, so idk if that is too much overhead

But if that's no problem for you I will look into it and implement it

SinanAkkoyun:
(the thing is that if I understand  this correctly:
```
❯ pip install guesslang                                                                                                           (exllama) 
Collecting guesslang
  Using cached guesslang-2.2.1-py3-none-any.whl (2.5 MB)
INFO: pip is looking at multiple versions of guesslang to determine which version is compatible with other requirements. This could take a while.
  Using cached guesslang-2.2.0-py3-none-any.whl (2.5 MB)
  Using cached guesslang-2.0.3-py3-none-any.whl (2.1 MB)
  Using cached guesslang-2.0.1-py3-none-any.whl (2.1 MB)
  Using cached guesslang-2.0.0-py3-none-any.whl (13.0 MB)
  Using cached guesslang-0.9.3-py3-none-any.whl (3.2 MB)
  Using cached guesslang-0.9.1-py3-none-any.whl (3.2 MB)
ERROR: Cannot install guesslang==0.9.1, guesslang==0.9.3, guesslang==2.0.0, guesslang==2.0.1, guesslang==2.0.3, guesslang==2.2.0 and guesslang==2.2.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    guesslang 2.2.1 depends on tensorflow==2.5.0
    guesslang 2.2.0 depends on tensorflow==2.5.0
    guesslang 2.0.3 depends on tensorflow==2.5.0
    guesslang 2.0.1 depends on tensorflow==2.2.0
    guesslang 2.0.0 depends on tensorflow==2.2.0
    guesslang 0.9.3 depends on tensorflow==1.7.0rc1
    guesslang 0.9.1 depends on tensorflow==1.1.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

```
❯ pip install tensorflow==2.5.0                                                                                                    (exllama) 
ERROR: Could not find a version that satisfies the requirement tensorflow==2.5.0 (from versions: 2.8.0rc0, 2.8.0rc1, 2.8.0, 2.8.1, 2.8.2, 2.8.3, 2.8.4, 2.9.0rc0, 2.9.0rc1, 2.9.0rc2, 2.9.0, 2.9.1, 2.9.2, 2.9.3, 2.10.0rc0, 2.10.0rc1, 2.10.0rc2, 2.10.0rc3, 2.10.0, 2.10.1, 2.11.0rc0, 2.11.0rc1, 2.11.0rc2, 2.11.0, 2.11.1, 2.12.0rc0, 2.12.0rc1, 2.12.0, 2.12.1, 2.13.0rc0, 2.13.0rc1, 2.13.0rc2, 2.13.0, 2.13.1, 2.14.0rc0, 2.14.0rc1, 2.14.0)
ERROR: No matching distribution found for tensorflow==2.5.0
```

there is no TF 2.5 anymore. So if I understand correctly (I am no dependency expert) it's non-trivial to install the required TF version, so that would be out of question I think)

turboderp:
Yeah, I wouldn't want to rely on TensorFlow in any case. Seems like kind of a large dependency for syntax highlighting.

It may make more sense to just have no syntax highlighting unless the model suggests a languate, with \`\`\`python or whatever. Or perhaps train a small micro-llama model for it. Idk.

SinanAkkoyun:
I also thought about a super tiny one layer model or something but in the end some good regex rules will rule I think. To be honest, the current langguessing really isnt too bad for all the scripting languages

I am just about done with the mistral fix, just give me a second before merging the other PR :D

turboderp:
I reworked it a bit cause I couldn't get it to behave. I think it's a little more stable now, and it only applies formatting once per line, so it's a bit faster. At least it recognizes the language now with Mistral-instruct, idk. Probably nothing you can really do if the model doesn't output a language, short of attaching another language model.

SinanAkkoyun:
Okay nice!

> it only applies formatting once per line

I tested it and noticed that it updates the whole block (coloring + block rendering) each line, which makes the block look laggy, in my original PR I only updated the lexguessing, assuming that was performance intensive.

It's just a super tiny detail and I will probably work on only coloring once per line (to keep your modifications performance boost) but updating the codeblock (black bg) + streamed text each streamed token when I find some time
(it's really a minor detail and no criticism, I just like the smoother codeblock gen)

> Probably nothing you can really do if the model doesn't output a language, short of attaching another language model.

What do you mean exactly? When it doesn't provide a codeblock at all or if the codeblock has no specified language? Tbh the current langguessing for me at least is sufficient enough so that it almost gets it right everytime

Thank you for improving on the syntax highlighting with me!

SinanAkkoyun:
Okay nice, thanks!

> it only applies formatting once per line

I tested it and noticed that it updates the whole block (coloring + block rendering) each line, which makes the block look laggy, in my original PR I only updated the lexguessing, assuming that was performance intensive.

It's just a super tiny detail and I will probably work on only coloring once per line (to keep your modifications performance boost) but updating the codeblock (black bg) + streamed text each streamed token when I find some time
(it's really a minor detail and no criticism, I just like the smoother codeblock gen)

> Probably nothing you can really do if the model doesn't output a language, short of attaching another language model.

What do you mean exactly? When it doesn't provide a codeblock at all or if the codeblock has no specified language? Tbh the current langguessing for me at least is sufficient enough so that it almost gets it right everytime

Thank you :)

turboderp:
It's mostly that it makes some very bad guesses, I feel. A lot of them still get *some* coloring, but it's inconsistent from one block to the next. That's almost worse than no syntax highlighting, when keywords are highlighted in comments, or when two versions of the same Python code are formatted as two different languages, neither of which is Python.

I'm tempted to start writing a more intelligent language guesser, since you could improve on it a lot by adding a few extra rules, e.g.:

- Lines starting with `# ` indicate Python, Bash, Ruby, YAML etc.
- Lines ending in `;` indicate C, Java, Go, Rust etc. 
- Lines starting in `let ` suggest JS, Rust, Haskell, etc. 
- f-strings are unique to Python (I think?)
- etc.

Of course those are also very straightforward cues for a tiny transformer to pick up on.

PR #74, opened 2023-09-30
Fixed Speculative Generator

SinanAkkoyun:
Hey! Just wanted to test the `speculative.py` example and noticed that the speculative generator did not use the (I suppose new) sampler syntax, so this is the fix for that. ^^

```py
# from
token, prob = ExLlamaV2Sampler.sample(logits, draft_settings, past, randoms[i])

# to:
token, prob, _ = ExLlamaV2Sampler.sample(logits, draft_settings, past, randoms[i], self.tokenizer)

# etc :)
```


SinanAkkoyun:
Ah something went wrong bc I used VS, let me create a new PR with only one file changed...

PR #75, opened 2023-09-30
Fixed Speculative Generator

SinanAkkoyun:
Hey! Just wanted to test the `speculative.py` example and noticed that the speculative generator did not use the (I suppose new) sampler syntax, so this is the fix for that. ^^

```py
# from
token, prob = ExLlamaV2Sampler.sample(logits, draft_settings, past, randoms[i])

# to:
token, prob, _ = ExLlamaV2Sampler.sample(logits, draft_settings, past, randoms[i], self.tokenizer)

# etc :)
```


turboderp:
Yeah, for token healing the sampler needs to call the tokenizer. So I missed that.

SinanAkkoyun:
I see, I'm so glad that you implemented speculative sampling as a nice generator, thank you!

PR #82, opened 2023-10-02
remake README+WS OPTIMIZE

Kerushii:


bkutasi:
Great work. I only have one thing to add: I just tested it with the latest Cuda 12.2, and it works without a problem. Maybe also adding a reference to the nvidia cuda toolkit docu would also help new users. 
https://docs.nvidia.com/cuda/

turboderp:
Yes, the JIT version should work for CUDA 12.2, there just isn't a prebuilt wheel for it.

PR #104, opened 2023-10-11
Check if Ampere GPUs or newer before using flash-attn

oobabooga:
This should solve the error

```
File "/miniconda3/envs/textgen/lib/python3.10/site-packages/flash_attn/flash_attn_interface.py", line 47, in _flash_attn_forward
    out, q, k, v, out_padded, softmax_lse, S_dmask, rng_state = flash_attn_cuda.fwd(
RuntimeError: FlashAttention only supports Ampere GPUs or newer.
```

that happens while trying to generate text if you have flash-attn installed (through a wheel) but do not have a compatible GPU.


turboderp:
Thanks. I hadn't considered people installing flash-attn without having hardware support for it.

PR #111, opened 2023-10-13
Use Pytorch 2.1 for CUDA 11.8+ and ROCm builds

jllllll:
For most systems, the wheels built on Pytorch 2.0.1 work fine with 2.1.
However, for unknown reasons, some systems seem to not work at all unless the wheel is built with Pytorch 2.1.

This PR changes the workflows to use the official Pytorch 2.1 release with CUDA 11.8+ and ROCm.
It also adds some basic error correction as I noticed that Mamba will not always report an error and stop the workflow if CUDA Toolkit fails to install.

I also commented out the PyPI upload functionality for non-sdist builds for convenience as it can't be used at the moment.
Should now only try to upload the JIT-compile wheel and the sdist itself.
I added a comment on line 179 detailing how to only upload the wheel and can add that change to this PR if you want.

Additionally, I made the following changes to the Powershell code that reads the package version number in the release workflows:
- Made the code more readable
- Now uses the `exllamav2/version.py` file for detection
- Now uses the release tag format of: `v{version_number}`

To clarify, it's purpose is to determine the package version number for uploading the wheels to a release using that version number.
It prevents automatic uploading to a release and will notify you in the event that the version number is not detected for some reason.

### DIFF
--- a/.github/workflows/build-wheels-release-rocm.yml
+++ b/.github/workflows/build-wheels-release-rocm.yml
@@ -1,11 +1,11 @@
-name: Build ROCm 5.4.2 Wheels & Release
+name: Build ROCm 5.6 Wheels & Release
 
 on:
   workflow_dispatch:
     inputs:
       pypi:
         description: 'Upload wheels to PyPI? 1 = yes, 0 = no'
-        default: '1'
+        default: '0'
         required: true
         type: string
   workflow_call:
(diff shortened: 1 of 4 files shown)


turboderp:
Thanks. I'm going to give these a try in a minute. :+1::+1:

jllllll:
Saw the error in your recent build.
From what I can tell, `cuda_fp8.h` is only present in CUDA 11.8.

turboderp:
Yep, I disabled it for 11.7 for now.

PR #112, opened 2023-10-14
support 8bit kv cache

zgce:
model.config.kv_cache_mask = '8bit' in use 8bit kv cache otherwize use 16bit kv cache


turboderp:
This is very interesting, and it seems to work alright. Saves a bunch of VRAM by only having one cache layer in FP16 at a time, and while I haven't measured perplexity or anything, it seems the output is fine...? At least fine. Hard to tell.

There is quite a heavy performance cost, though. About a 50% - 70% drop in speed. I guess some of it could be mitigated by not converting the whole FP8 cache layer to FP16, and only converting back the updated keys/values. I'd also want to profile the kernels to see how efficient the conversion is.

zgce:
My idea is just to save KV Cache VRAM, because running LLM well on many poor graphics cards is only a few gigabytes of memory. 

It is certainly a better idea to convert only the updated key values, but I have no idea how to identify which key values have been updated.

turboderp:
Well, I've included it as an "undocumented feature" for now. Thanks so much for the work, it'll definitely be useful for cramming larger models into smaller GPUs. I'll do some profiling in a bit and see about limiting the FP8-FP16 conversions.

PR #121, opened 2023-10-19
Fix some syntax errors

nivibilla:
Hi, just found a small error in the lora example


turboderp:
Yeah, Python is so forgiving :(

PR #137, opened 2023-10-31
Fix the garbadge output for ROCM

deltaguo:
Fix the garbadge output for rocm5.6, gfx90a.
The kernel of rms_norm uses `__shfl_xor_sync()` for reduce. There are 64 threads in ROCM warp, while 32 in cuda. Using a 32x32 layout in each block will cause unexpected results. Changing it to 64x16 solved this problem.


fgdfgfthgr-fox:
Can confirm this solve the issue for me.

fgdfgfthgr-fox:
Wait... this seems to create a new issue. After applying this fix, I can no longer quantize model. It will say Hessian is not invertible when trying to quant.

turboderp:
If the problem is with the RMS norm kernel, what happens if you just force the use of `forward_torch()` in rmsnorm.py? That's a plain Torch implementation that.

Jipok:
`python examples/inference.py`
```
Loading model: /home/llama/models/Mistral-7B-instruct-exl2/
  File "/home/llama/exllamav2-fix/examples/inference.py", line 30, in <module>
    cache = ExLlamaV2Cache(model, lazy = True)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ExLlamaV2Cache.__init__() got an unexpected keyword argument 'lazy'
```
`python test_inference.py -m ~/models/Mistral-7B-instruct-exl2 -p "Once upon a time,"`
```
 -- Model: /home/llama/models/Mistral-7B-instruct-exl2
 -- Options: ['rope_scale 1.0', 'rope_alpha 1.0']
 -- Loading model...
 -- Loading tokenizer...
 -- Warmup...
 -- Generating...

Once upon a time,125040080130979623500808910013223301204400082530023291248043012401.233892337101200140190300
22002040423019430233094014401033016

 -- Response generated in 3.03 seconds, 128 tokens, 42.26 tokens/second (includes prompt eval.)
```
M.b. i need download new model?

turboderp:
Do all ROCm devices have a warp size of 64?

deltaguo:
so far yes

ardfork:
> Do all ROCm devices have a warp size of 64?

No, most consumer one use 32, only pro one use 64, my GPU use 32 for example. PyTorch handle it that way: https://github.com/pytorch/pytorch/blob/d7f698102e5b889af09082b86d23b9407bfe6863/c10/macros/Macros.h#L323-L327

I think you might even be able to use `-mno-wavefrontsize64` to force it to 32 on pro GPU.

turboderp:
I guess I can work with `warpSize` then. I'll get on it.

fgdfgfthgr-fox:
```
(textgen) fgdfgfthgr@fgdfgfthgr-MS-7C95:/mnt/7018F20D48B6C548/exllamav2$ python convert.py -i "/mnt/7018F20D48B6C548/exllamav2/yi_34b" -o "/mnt/7018F20D48B6C548/exllamav2/convert_output" -c "/mnt/7018F20D48B6C548/exllamav2/calibration_data_file/0000.parquet" -l 4096 -b 3.5
 !! Warning: calibration rows > 2048 tokens may result in excessive VRAM use
 -- Beginning new job
 -- Input: /mnt/7018F20D48B6C548/exllamav2/yi_34b
 -- Output: /mnt/7018F20D48B6C548/exllamav2/convert_output
 -- Calibration dataset: /mnt/7018F20D48B6C548/exllamav2/calibration_data_file/0000.parquet, 100 / 16 (0) rows, 4096 tokens per sample
 -- Target bits per weight: 3.5 (decoder), 6 (head)
 -- Max shard size: 8192 MB
 -- RoPE scale: 1.00
 -- RoPE alpha: 1.00
 -- Tokenizing samples (measurement)...
 -- First 50 tokens of dataset:
    ' = Valkyria Chronicles III = \n Senjō no Valkyria 3 : Unrecorded Chronicles ( Japanese : 戦場のヴァルキュリア3'
 -- Last 50 tokens of dataset:
    ' a love @-@ struck Barbie . " Critic Anupama Chopra also criticized Fernandez , calling her " a pin @-@ prick on a balloon " . Later that year , she made a cameo'
 -- Token embeddings (measurement)...
 -- Measuring quantization impact...
 -- Layer: model.layers.0 (Attention)
 -- Linear: model.layers.0.self_attn.q_proj
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
 !! Warning: Applied additional damping
  File "/mnt/7018F20D48B6C548/exllamav2/conversion/adaptivegptq.py", line 239, in prepare
    hessian_inv = torch.linalg.cholesky(hessian)
torch._C._LinAlgError: linalg.cholesky: The factorization could not be completed because the input is not positive-definite (the leading minor of order 1 is not positive-definite).

During handling of the above exception, another exception occurred:

  File "/mnt/7018F20D48B6C548/exllamav2/convert.py", line 270, in <module>
    measure_quant(job, save_job, model)
  File "/mnt/7018F20D48B6C548/exllamav2/conversion/quantize.py", line 265, in measure_quant
    test_quants(module.q_proj, quantizers["q_proj"], post_norm, query_states, qparams_options, results)
  File "/mnt/7018F20D48B6C548/exllamav2/conversion/quantize.py", line 107, in test_quants
    if not skip_prep: lq.prepare()
  File "/mnt/7018F20D48B6C548/exllamav2/conversion/adaptivegptq.py", line 270, in prepare
    raise ValueError("Hessian is not invertible")
ValueError: Hessian is not invertible
```
This error only happens if applying this pull request. Else it quant normally.

turboderp:
@fgdfgfthgr-fox Is this on a warp size 32 device?

fgdfgfthgr-fox:
> @fgdfgfthgr-fox Is this on a warp size 32 device?

It's on a radeon vii, which I suppose should be warp size 64.

turboderp:
Okay, so I do want to merge this, and sorry for being a little obtuse maybe. But I just need clarification since I have no way to test this myself and I don't want to break anything for ROCm users who currently aren't having issues. Is there any reason not to do this?

```c++
#define WARP_SIZE (warpSize)
#define NUM_WARPS (1024 / warpSize)
```

IMbackK:
yes that would work correctly

IMbackK:
@turboderp  i can confirm that that dose fix the issue on gfx900 and gfx906, but i dont think you should be doing this on the hip side only in a define warpSize is a property hip inherits from cuda https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp_18656f53eb2a7e54500f6fb95a830b47d and just assuming 32 just because thats true for current devices is not really valid cuda either.

PR #146, opened 2023-11-07
response id and infer modification

Kerushii:
one request id usually corresponds to one response id; for that, request ids and response ids are mostly the same thing in most cmds, however, in the case of stream, one request will yield multiple responses, so a client provided response id doesnt make sense: different responses fired using the same response id doesn't make sense. therefore, in the case of multi firing cmds, only the first response will have a response id. all subsequently triggered response by the same request will have the same request ID

### DIFF
--- a/exllamav2/exllamav2_ext/cuda/rope.cu
+++ b/exllamav2/exllamav2_ext/cuda/rope.cu
@@ -121,7 +121,7 @@ void rope_cuda
     dim3 blockDim, gridDim;
     blockDim.x = THREADS_X;
     blockDim.y = threads_y;
-    gridDim.x = DIVIDE(head_dim, THREADS_X) / (use_half2 ? 2 : 1);
+    gridDim.x = DIVIDE(head_dim / (use_half2 ? 2 : 1), THREADS_X);
     gridDim.y = DIVIDE(rows_per_batch, threads_y);
     gridDim.z = batch_size;
 


turboderp:
>therefore, in the case of multi firing cmds, only the first response will have a response id. all subsequently triggered response by the same request will have the same request ID

I don't understand the reasoning here. The code explicitly disallows having both request and response ID in the same client request, so this would result in one reply that has a response ID but no request ID, and then the rest of the streamed response would have neither?

Isn't it much cleaner for the client to send a request, then distinguish by the `response_type` field whether it's receiving one of a series of chunks or the full response once the response is complete? I.e. the client sends:

`{ action: "infer", request_id: "12345", text: "Once upon a time,", max_new_tokens: 1000, stream: True, tag: "story_time"}`

Then the server sends:

`{ action: "infer", request_id: "12345", response_type: "chunk", chunk: " there", tag: "story_time" }`
`{ action: "infer", request_id: "12345", response_type: "chunk", chunk: " lived", tag: "story_time" }`
`{ action: "infer", request_id: "12345", response_type: "chunk", chunk: " a", tag: "story_time" }`
`{ action: "infer", request_id: "12345", response_type: "chunk", chunk: " princ", tag: "story_time" }`
`{ action: "infer", request_id: "12345", response_type: "chunk", chunk: "ess", tag: "story_time" }`
`...`
`{ action: "infer", request_id: "12345", response_type: "full", util_text: "Once upon a time,", response: " there lived a princess in a castle on top of a...", tag: "story_time" }`

If the problem is distinguishing between the responses so that the first one is unique, it seems much cleaner to just add a serial number:

```python
    completion = ""
    gen_tokens = 0
    chunk_id = 0

    while True:
        chunk, eos, _ = server.generator.stream()
        completion += chunk
        gen_tokens += 1

        if stream and chunk != "":
            response["response_type"] = "chunk"
            response["chunk"] = chunk
            response["chunk_id"] = chunk_id
            chunk_id += 1
            await ws.send(json.dumps(response))

        if eos or gen_tokens >= num_tokens: break
```

This wouldn't be a breaking change, and client-side you could simply do something like:

```javascript
if (response.chunk_id == 0)
    response.response_id = "cat";
else
    response.response_id = "dog";
```

Or whatever else to distinguish.

Also, streaming the entire response-so-far with every token seems excessive. That should at the very least be optional since it creates potentially megabytes of traffic to stream kilobytes of text. I know it's 2023, but that still seems like a lot.

Issue #46, opened 2023-09-19
Ninja Build Error for ROCm


lufixSch:
I get a bunch of errors when running.

```bash
python test_inference.py -m <model> -p "Once upon a time,"
```

On the first execution the process exited with the error message provided in the following File. The Output is pretty long therefore I uploaded it as a separate file

[Output of the first run](https://github.com/turboderp/exllamav2/files/12659904/logs_old.txt) 

After the first execution I noticed, that the output changed. Now I get this Error:

[Output of the second run](https://github.com/turboderp/exllamav2/files/12659919/logs.txt)

I am not sure If there are some extra steps necessary when running ExllamaV2 on ROCm. I just cloned the repository and installed the requirements as stated in the README.md (torch for ROCm was installed before).

### System Information:
Operating System: Manjaro Linux 
KDE Plasma Version: 5.27.7
KDE Frameworks Version: 5.109.0
Qt Version: 5.15.10
Kernel Version: 6.5.1-1-MANJARO (64-bit)
Graphics Platform: X11
Processors: 20 × 13th Gen Intel® Core™ i5-13500
Memory: 31,1 GiB of RAM
Graphics Processor: AMD Radeon RX 6750 XT
Manufacturer: Micro-Star International Co., Ltd.
Product Name: MS-7D98
System Version: 1.0

ardfork:
It seems like you are missing hipBLAS.

lufixSch:
When I check with `pacman` it lists a `hipblas` package

```bash
$ pacman -Q hipblas  
hipblas 5.6.0-1
```

lufixSch:
Pytorch is installed via pip. The venv I am using is the one I am running Oobaboogas WebUI in.

Thanks, I will try that patch later.

lufixSch:
I tried the patch but it did not work. The Error is still the same.

Ps.: I was unable to apply the patch, therefore I changed the file manually

I tried searching for the `hipblas` library with ldconfig and got the following results:

```bash
$ ldconfig -p | grep hipblas    
libhipblas.so.1 (libc6,x86-64) => /opt/rocm/lib/libhipblas.so.1
libhipblas.so (libc6,x86-64) => /opt/rocm/hipblas/lib/libhipblas.so
libhipblas.so (libc6,x86-64) => /opt/rocm/lib/libhipblas.so
```

Not sure if that helps.

ardfork:
[...]

PR #22, opened 2023-09-13
Fix Compiling with HIP on Older Pytorch Version

leonxia1018:
Hit a compiling issue when running the model with Rocm.
```
Traceback (most recent call last):
  File "/opt/conda/envs/py_3.9/lib/python3.9/site-packages/torch/utils/cpp_extension.py", line 1900, in _run_ninja_build
    subprocess.run(
  File "/opt/conda/envs/py_3.9/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['ninja', '-v', '-j', '8']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/lib/jenkins/text-generation-webui/repositories/exllamav2/test_inference.py", line 2, in <module>
    from exllamav2 import(
  File "/var/lib/jenkins/text-generation-webui/repositories/exllamav2/exllamav2/__init__.py", line 3, in <module>
    from exllamav2.model import ExLlamaV2
  File "/var/lib/jenkins/text-generation-webui/repositories/exllamav2/exllamav2/model.py", line 12, in <module>
    from exllamav2.linear import ExLlamaV2Linear
  File "/var/lib/jenkins/text-generation-webui/repositories/exllamav2/exllamav2/linear.py", line 4, in <module>
    from exllamav2 import ext
  File "/var/lib/jenkins/text-generation-webui/repositories/exllamav2/exllamav2/ext.py", line 118, in <module>
    exllamav2_ext = load \
  File "/opt/conda/envs/py_3.9/lib/python3.9/site-packages/torch/utils/cpp_extension.py", line 1283, in load
    return _jit_compile(
  File "/opt/conda/envs/py_3.9/lib/python3.9/site-packages/torch/utils/cpp_extension.py", line 1508, in _jit_compile
    _write_ninja_file_and_build_library(
  File "/opt/conda/envs/py_3.9/lib/python3.9/site-packages/torch/utils/cpp_extension.py", line 1623, in _write_ninja_file_and_build_library
    _run_ninja_build(
  File "/opt/conda/envs/py_3.9/lib/python3.9/site-packages/torch/utils/cpp_extension.py", line 1916, in _run_ninja_build
    raise RuntimeError(message) from e
[...]

PR #154, opened 2023-11-09
Simplify HIP compatibility

ardfork:
Since #22 didn't seem to fix using exllamav2 on ROCm < 5.6 as reported on #46. I think it's better to use `HIPBLAS_USE_HIP_HALF` which require ROCm 5.5.0; that will make less code used only for HIP compatibility.


setup.py
```diff
@@ -15,9 +16,14 @@
 if ext_debug:
     extra_cflags += ["-ftime-report", "-DTORCH_USE_CUDA_DSA"]
 
+extra_cuda_cflags = ["-lineinfo", "-O3"]
+
+if torch_version.hip:
+    extra_cuda_cflags = ["-DHIPBLAS_USE_HIP_HALF"]
```
turboderp, line 22: Should this be `+=` ?

turboderp:
This makes sense, though I have no way to test it. Is there a performance impact?

setup.py
ardfork: Should have been, fixed.

ardfork:
> Is there a performance impact?

Not that I could notice, benchmark results are identical on this PR or master.

turboderp:
Is it backwards-compatible? I.e. do I just change the build environment to `ROCM_VERSION=5.5`  and it should work for people still running 5.6?

ardfork:
Huh, I think you should have let it to 5.6, that what most distro use and what PyTorch use. 5.5 is just when `HIPBLAS_USE_HIP_HALF` was introduced.

turboderp:
Ooh..  I thought it was about something being deprecated in 5.6. I'll just revert then.

ardfork:
No, originally that code came from exllama v1 where latest was 5.4 at the time, I believe. Then came exllama v2 and I copied hipblas compatibility code without knowing that `HIPBLAS_USE_HIP_HALF` was introduced.

ardfork:
Also, I wanted to modify README, to maybe add some ROCm instruction (adding `--extra-index-url https://download.pytorch.org/whl/rocm5.6` to pip command), but I don't think the instruction even work if you don't already have pytorch installed. `python setup.py install --user` just fail since it try to import stuff from torch.

turboderp:
I guess PyTorch ran into the same issue I did, being that the main PyPi index doesn't support multiple variants of one package. I'm not sure what the best approach is, though. Should there be one `requirements.txt` for every CUDA or ROCm version, split into Windows and Linux..? I guess that only works out to six files at the moment.

But as long as you have *a* version of PyTorch >=2.1.0, the requirement is satisfied, at least.

PR #189, opened 2023-11-28
Add Huggingface tokenzier support

DOGEwbx:
Add logic to decide use huggingface tokenizer or sentence piece tokenizer.
It can support models using huggingface tokenizer like Falcon and Deepseek Coder

### DIFF
--- a/exllamav2/model_init.py
+++ b/exllamav2/model_init.py
@@ -43,14 +43,18 @@ def check_args(args):
         sys.exit()
 
     required_files = ["config.json",
-                      "tokenizer.model",
+                      ["tokenizer.model", "tokenizer.json"],
                       "*.safetensors"]
 
     for filename in required_files:
-
-        path = os.path.join(args.model_dir, filename)
-        matches = glob.glob(path)
-        if len(matches) == 0:
+        if isinstance(filename, str):
+            filename = [filename]
+        all_matches = []
+        for file in filename:
+            path = os.path.join(args.model_dir, file)
+            matches = glob.glob(path)
+            all_matches += matches
+        if len(all_matches) == 0:
             print(f" ## Error: Cannot find {filename} in {args.model_dir}")
             sys.exit()
 
--- a/exllamav2/tokenizer.py
+++ b/exllamav2/tokenizer.py
@@ -1,7 +1,108 @@
 from exllamav2.config import ExLlamaV2Config
 from sentencepiece import SentencePieceProcessor
+from transformers import AutoTokenizer
+from typing import List, Union
 import torch
 import os, json, re
+import warnings
+
+class BaseTokenizer:
+    def __init__(self, model) -> None:
+        self.tokenizer_model = model
+
+    def unk_id(self) -> int:
+        return self.tokenizer_model.unk_id()
+    
+    def pad_id(self) -> int:
+        return self.tokenizer_model.pad_id()
+    
+    def pad_token(self) -> str:
+        return None
+    
+    def bos_id(self) -> int:
+        return self.tokenizer_model.bos_id()
+    
+    def bos_token(self) -> str:
+        return None
+
+    def eos_id(self) -> int:
+        return self.tokenizer_model.eos_id()
+    
+    def eos_token(self) -> str:
+        return None
+    
+    def vocab_size(self) -> int:
+        return self.tokenizer_model.vocab_size()
+    
+    def id_to_piece(self, idx: Union[int, List[int]]) -> str:
+        return self.tokenizer_model.id_to_piece(idx)
+    
+    def decode(self, idx: int) -> str:
+        return self.tokenizer_model.decode(idx)
+    
+    def Decode(self, ids: List[int]) -> str:
+        return self.tokenizer_model.Decode(ids)
+    
+    def Encode(self, text: str) -> list:
+        return self.tokenizer_model.Encode(text)
+    
+    def EncodeAsIds(self, text: str) -> list:
+        return self.tokenizer_model.EncodeAsIds(text)
+
+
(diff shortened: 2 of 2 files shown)


CyberTimon:
Thank you so much for this work @DOGEwbx . I'm waiting for deepseek coder support all the time. This is very helpful.

CyberTimon:
Is creating Exl2 quants with this also possible?

DOGEwbx:
@CyberTimon Thanks for your interest on our work. I haven’t run testing on exl2 quants file but as all the modifications are on the tokenizer part, I don't think there will be problems on the specific data format.
I'm not very familiar with model quant techniques, if you could give me some model checkpoints or conversion scripts so that I can do some tests.

turboderp:
At the very least, this is going to take some time to review.

Transformers is a *massive* dependency to include just to support one model (Falcon still wouldn't work as there are other architectural differences).

As for remote code, my guess would be that 90% of users are unaware of the risks involved, so it should at least be opt-in.

I'll need a moment to think about it, to test that this doesn't break functionality like constrained sampling, and make sure there really isn't a better way.

DOGEwbx:
Thanks for your reply.
For the transformers dependency issue, use the Tokenizers module instead would be a solution.
And you're right for trust_remote_code because I don't know the risks either.....We could disable it since the model should be on local.

SinanAkkoyun:
Is there any specific way to use the fork? With `pip install transformers` the fork does not work for me:

```
python examples/chat.py -m../../models/deepseek/deepseek-coder-1.3b-instruct-GPTQ/ -mode deepseek
 -- Model: ../../models/deepseek/deepseek-coder-1.3b-instruct-GPTQ/
 -- Options: ['rope_scale 1.0', 'rope_alpha 1.0']
 -- Loading model...
 -- Loading tokenizer...
 -- Prompt format: deepseek
 -- System prompt:

You are an AI programming assistant, utilizing the Deepseek Coder model, developed by Deepseek Company, and you only answer questions related to computer science. For politically sensitive questions, security and privacy issues, and other non-computer science questions, you will refuse to answer.

User: How can I tell the time in python?

���ııiiiii iii iv i i ii io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io io
```

6.7B shows very similar behaviour, but often returns some kind of invisible token in an infinite loop

I get the same behaviour no matter what prompt format ( [the prompt format](https://github.com/SinanAkkoyun/exllama-v2/tree/deepseek) )

Maybe I am just doing something wrong, I'd appreciate help

turboderp:
@SinanAkkoyun The model seems to use linear a RoPE scaling factor of 4. I've been able to get coherent output out of the 1.3B model at least, using that.

@DOGEwbx The Tokenizers library seems like a more reasonable dependency, especially if it's optional. It largely mirrors Transformers, so it should be possible to adapt it to the code in this PR. There are still a few things I need to sort out and verify, like how control symbols are encoded, optional BOS/EOS tokens, that the vocabulary is preprocessed correctly, how UTF-8 characters are emitted and so on. I'll get to that in a few hours.

It's definitely not a trivial issue. I see over on the llama.cpp repo a whole bunch of people have been working on it for some weeks now.

As for remote code, the issue is that with the option enabled, `AutoModel.from_pretrained` and apparently also `AutoTokenizer.from_pretrained` will import and run architectures distributed with models. The architectures are defined as Python code, and with no sandboxing this code has access to your entire userspace. So it can steal browser cookies, take screenshots, read the clipboard, etc. Crucially, users usually don't think of downloading and trying out a new model as downloading and installing software, especially in UIs like text-generation-webui that let you download and run models with a couple of clicks.

SinanAkkoyun:
@turboderp Thank youu, 6.7B is also working coherently :)

turboderp:
I think maybe you're just asking too much of a tiny model. And quantization is known to affect smaller models more severely anyway.

turboderp:
There. I rewrote it to use the Tokenizers library instead, as an optional dependency, and it seems to run okay now. It seems to consistently encode and decode the same as a HF AutoTokenizer. Encoding seems to work correctly during quantization as well.

I also added a workaround for the Tokenizer bug where some added tokens would decode incorrectly. Still need to test it with some of the other models that lack a SentencePiece tokenizer model.

CyberTimon:
Thank you

Majid121232 (approved):

Majid121232 (approved):

PR #191, opened 2023-11-28
Fixed code block syntax highlighting for long code

SinanAkkoyun:
When very long codeblocks are being printed, the ANSI clear line code can not clear the whole code to reprint, because it is outside of the terminal window.

Before, the codeblock repeated it's top lines as they were not cleared but still printed.

As a fix, only clear termheight-1 lines, the remaining top line should automatically stay in buffer.

(The PR works but the top line somehow messes up the front padding, I am working on that rn)


SinanAkkoyun:
The padding now also works

PR #217, opened 2023-12-07
[Draft] Trying to add QuiP quant into inference

waters222:
This is draft PR for adding QuiP quant into ExllamaV2.
[Original QuiP Repo](https://github.com/Cornell-RelaxML/quip-sharp)

Works:
1. right it it can load [pre-quant model](https://huggingface.co/relaxml/Llama-2-13b-chat-E8P-2Bit) into model and generate token.
2. since origin model does not have tokenizer.model file etc. so it needs to copy HF model file into this.

Does not works:
1. The VRAM consumption is almost twice the intend size. 
2. The generated token looks giberish. 

Example output
```
Loading model: /media/storage/models/Llama-2-13b-chat-E8P-2Bit
once upon a time hex horn offici dry Dutch date inspect natur Speciesnr offici accident lod exped Á Sever offic offici permanéri sé perman inspect Oktober Oktober perman cat Hell Áériław lod offici periodic Hell Á lifetime Á Sever perman inspect offici Sever circa naturław offic succession circaween

Response generated in 3.25 seconds, 50 tokens, 15.37 tokens/second
```

What I was hopping is that @turboderp  can help me fix both the VRAM issue and generation problem.
I think the cause maybe in the attention layer.


CyberTimon:
Very cool. Thank you for making this, will be awesome when it's done 👍

KnutJaegersberg:
Would like to have this feature :)

tau0-deltav:
It might be worth nothing that most of the what QuIP# appears to be achieving seems to be acheived better by llama.cpp's new quantization schema. the dedicated 2 bit quants (IQ2_XS and XXS, 2.03 and 2.3 BPW - as opposed to being more like 3 bits) are very strong.* 

There's now an optimisation scheme involved but it's clearly not a 'generate FP64 hessians for a week (no really that's what they suggested - for 6k ctx) on your grandma's Threadripper X cluster' - it's much more like the 'discover the most important weights by throwing words at the decoder' - schema we're familiar with here.

[ikawrakow](https://github.com/ikawrakow) [here is well worth a read.](https://github.com/ggerganov/llama.cpp/pull/4773#issue-2065647760)  I'll link this technical discussion I found instead of the unsightly spat with one of the Cornell team.

I'm long past a claim to being any sort of computer scientist but I'd like to hope EXL2's inheritance from GPTQ (quantize-then-calibrate as GPTQ'sformat design goal, flexibility in weight assignments added by EXL2 from that) could make EXL2 itself a better use for the methods used here than QuIP#. Those imatrix files are bigger than exl2's measurements but 25mb isn't exactly out of reach here, compared to 2mb?

(...Could these just convert directly? Probably no without the E8  enumeration but q

*IQ3_XXS is Absolutely Robust it's scary.

PR #218, opened 2023-12-08
Fixed multi file and wildcard args

SinanAkkoyun:
Before it couldn't handle wildcard args for me

### DIFF
--- a/util/convert_safetensors.py
+++ b/util/convert_safetensors.py
@@ -2,29 +2,14 @@ import torch
 import argparse, os, glob, sys
 from safetensors.torch import save_file
 
-parser = argparse.ArgumentParser(description = "Convert .bin/.pt files to .safetensors")
-parser.add_argument("input_files", type = str, help = "Input file(s)")
+parser = argparse.ArgumentParser(description="Convert .bin/.pt files to .safetensors")
+parser.add_argument("input_files", nargs='+', type=str, help="Input file(s)")
 args = parser.parse_args()
 
-tensor_file_pattern = args.input_files
-tensor_files = glob.glob(tensor_file_pattern)
-
-if len(tensor_files) == 0:
-    print(f" ## No files matching {tensor_file_pattern}")
-    sys.exit()
-
-for file in tensor_files:
+for file in args.input_files:
     print(f" -- Loading {file}...")
-    state_dict = torch.load(file, map_location = "cpu")
+    state_dict = torch.load(file, map_location="cpu")
 
     out_file = os.path.splitext(file)[0] + ".safetensors"
     print(f" -- Saving {out_file}...")
     save_file(state_dict, out_file)
-
-
-
-
-
-
-
-


bartowski1182:
@SinanAkkoyun can you provide an example of inputting wildcards? This change seems to have broken wildcard input for me

bartowski1182:
Reverting this commit fixes wildcard input for me:

python3 utils/convert_safetensors.py /models/model_name/pytorch*.bin

turboderp:
Interesting. What shell are you using?

bartowski1182:
Just Linux bash

bartowski1182:
@SinanAkkoyun any info on how you're using this? I'm still reverting this merge for my use case because I can't convert with wildcards anymore.

bartowski1182:
@turboderp do you mind reverting this commit? If this was fixing a real issue I can look at a better fix but for now it's frustrating that it broke wildcard support

SinanAkkoyun:
@bartowski1182 That's odd, for me it works with simply specifying: `python3 utils/convert_safetensors.py /models/model_name/pytorch*` (I can not try it out now, I do not have access to my machine atm)

It seems as if the PR only breaks for your environment, if that's the case you could simply copy the old util script into another directory and run it there to convert

bartowski1182:
@SinanAkkoyun is that how you would use it? I can't imagine how it would work with a wildcard from the code, since you just run for file in args.input_files:, that doesn't seem to handle using a wildcard at all but would work with a list of files

SinanAkkoyun:
Idk how but my PR worked for me, perhaps my shell handled that

SinanAkkoyun:
I can retry in a couple of weeks when I get access to my machine

SinanAkkoyun:
Just set a simplified version up on the go, tested the input file arg parsing, works fine for fish and bash (ubuntu 20.04)

with * at the end, with *.bin at the end, all work flawlessly

turboderp:
It also works for me (bash). Previously you'd have to run it like `python convert_safetensors.py "*.bin"` since otherwise `*.bin` would be expanded by the shell to a list of files matching the pattern. Adding `nargs='+'` correctly catches that list and so wildcard matching can be done by the shell.

What happens if you run this?

```python
import argparse
parser = argparse.ArgumentParser()
parser.add_argument("filenames", nargs="+", type=str)
args = parser.parse_args()
print(args.filenames)
```

bartowski1182:
I've isolated the problem a bit

If I run the code directly from my terminal, it works

If I run it inside a docker container, it doesn't..

I'm not sure how running from a docker container is making it not work, it doesn't make any sense to behave differently, but now that I know what I'm looking for I should have an easier time either figuring out what's broken or figuring out how to fix this code so that it works in both scenarios

bartowski1182:
Made a fix for it, seems that docker commands won't expand wildcard files like regular bash does, will now work for both:

https://github.com/turboderp/exllamav2/pull/344

PR #221, opened 2023-12-09
Support DragonFox style "BaNnbAnN"

Kerushii:
I will BaNnbAnNbA</s>

### DIFF
--- a/exllamav2/server/websocket_actions.py
+++ b/exllamav2/server/websocket_actions.py
@@ -142,6 +142,11 @@ async def infer(request, ws, server, response):
         if not isinstance(ss, list): ss = [ss]
         sc += ss
 
+    if "bann_bann" in request:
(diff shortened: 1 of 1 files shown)


BetaDoggo:
Skibidiers

exllamav2/server/websocket_actions.py
```diff
@@ -167,6 +172,7 @@ async def infer(request, ws, server, response):
     gs.typical = float(request["typical"]) if "typical" in request else 0
     gs.temperature = float(request["temperature"]) if "temperature" in request else 0.9
     gs.token_repetition_penalty = float(request["rep_pen"]) if "rep_pen" in request else 1.05
+    gs.disallow_tokens(server.tokenizer, bb)
```
turboderp, line 175: Wouldn't this break if there's no "bann_bann" key in the request?

PR #224, opened 2023-12-12
Stop conditions and exclude prompt for Base generator

SinanAkkoyun:
I needed the base gen to work with custom stop conditions (mainly for batched inference reasons).
Now, stop_conditions (str list) and multiple stop_tokens (compatible with supplying an ID instead of ID list) are implemented

I also added an optional param to exclude the prompts from the output


turboderp:
This should already be implemented, and in any case the dynamic generator has a unified interface that allows for it.

SinanAkkoyun:
Back then it didn't iirc, thanks for your work!

PR #237, opened 2023-12-21
add flash attention feature to different seqlen batch

fahadh4ilyas:


fahadh4ilyas:
I just realized it contains copyrighted code from huggingface so I have to change it.

PR #263, opened 2024-01-06
Add dynatemp (the entropy one)

awtrisk:
Still some stuff to be checked, heavy wip.


awtrisk:
alright, i'm not sure - are the actual logits `temp_probs`?

turboderp:
The `sample_basic` function takes the input logits and converts them to a probability distribution (calling `softmax_cpu`). The result is `temp_probs` and `temp_indices` which are processed by the subsequent samplers. I have no idea if it's more correct to apply this dynamic temperature before or after the other samplers.

bdashore3:
Looking at the newer implementation in backends such as kcpp. It's definitely better to pass a single dynatemp_range parameter into the temperature sampler instead of overburdening the function with min, max, and enabled values. I believe the WIP code here should be refactored to reflect that. The frontend/API can handle calculation of the range to pass to exl2.

bdashore3:
Now that dynatemp has matured more, I'd like to stick to an implementation which will be inline with other backends such as Aphrodite.

1. Merge dynatemp into the temp function (to allow temp_last to work)
2. Remove dynatemp_enabled which is a redundant boolean
3. Add a check within the temp function to enable dynatemp based on the min and max values along with if max > min.
4. Add the option to use dynatemp_exponent to adjust that as well.

CC: @turboderp

awtrisk:
> Now that dynatemp has matured more, I'd like to stick to an implementation which will be inline with other backends such as Aphrodite.
> 
>     1. Merge dynatemp into the temp function (to allow temp_last to work)
> 
>     2. Remove dynatemp_enabled which is a redundant boolean
> 
>     3. Add a check within the temp function to enable dynatemp based on the min and max values along with if max > min.
> 
>     4. Add the option to use dynatemp_exponent to adjust that as well.
> 
> 
> CC: @turboderp

All done - dynatemp is in the temp function, and the exponent can be adjusted.

bdashore3:
If you believe the PR is okay to review. Please remove draft status.

PR #275, opened 2024-01-12
Repeat layers to create FrankenModels

dnhkng:
## Description
This slightly modifies the forward pass, to reuse layers to allow the creation and use of 'Frankenmodels' quickly and easily.

The format of the new argument is:
`python test_inference.py  -m /models/lzlv_70b_fp16_hf-4.0bpw-h6-exl2 -p "Once upon a time:" -gs 18,18 --repeats '[(0,20),(10,30),(20,40),(30,50),(40,60),(50,70),(60,79)]'`

This would generate the [nsfwthrowitaway69/Venus-120b-v1.2
](https://huggingface.co/nsfwthrowitaway69/Venus-120b-v1.2/blob/main/mergekit_config.yml) Frankenmodels, while reducing VRAM usage by 50b params.

The repeats parameter is a string list of tuples. As the final layers in most models are model.norm and lm_head (not a number), the last value in the last tuple should be one lower than the final layer number. So long as this is the case, the code will extend out the layer so all are included. i.e. `[(0,20),(10,28)] and  [(0,20),(10,30)]` would generate the same Frankenmodel.

## Related Discussion
discussion at #270 and a discussion on Reddit [localllama](https://www.reddit.com/r/LocalLLaMA/comments/190zsk1/comment/kgu7ffj/) about the potential of easily creating Frankenstein models using exllama.

## Explanation of changes
A new parameter was added to argparse, and in ExLlamaV2.__init__ if the param was used, we build a list of the layer order to use including repeats.  In the ExLlamaV2._forward method, the actual model forward pass is extracted into a private process_module method, and this is called in the usual way with looping through self.modules if 'repeats' is not passed, and looping through self.layers_list if it is passed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__/
 .idea
 venv
 dist
+*.so
--- a/exllamav2/model.py
+++ b/exllamav2/model.py
@@ -130,14 +130,14 @@ class ExLlamaV2:
         self.modules.append(ExLlamaV2Embedding(self, "model.embed_tokens"))
         self.modules_dict[self.modules[-1].key] = self.modules[-1]
 
-        for layer_idx in range(self.config.num_hidden_layers):
+        for layer_list in range(self.config.num_hidden_layers):
 
-            self.modules.append(ExLlamaV2Attention(self, f"model.layers.{layer_idx}", layer_idx))
+            self.modules.append(ExLlamaV2Attention(self, f"model.layers.{layer_list}", layer_list))
             for m in self.modules[-1].submodules: self.modules_dict[m.key] = m
             if self.config.architecture == "Mixtral":
-                self.modules.append(ExLlamaV2MoEMLP(self, f"model.layers.{layer_idx}", layer_idx))
+                self.modules.append(ExLlamaV2MoEMLP(self, f"model.layers.{layer_list}", layer_list))
             else:
-                self.modules.append(ExLlamaV2MLP(self, f"model.layers.{layer_idx}", layer_idx))
+                self.modules.append(ExLlamaV2MLP(self, f"model.layers.{layer_list}", layer_list))
             for m in self.modules[-1].submodules: self.modules_dict[m.key] = m
 
 
@@ -150,14 +150,39 @@ class ExLlamaV2:
 
         # Find last layer that affects k/v cache
 
-        layer_idx = len(self.modules)
+        layer_list = len(self.modules)
         while True:
-            layer_idx -= 1
-            if isinstance(self.modules[layer_idx], ExLlamaV2Attention):
+            layer_list -= 1
+            if isinstance(self.modules[layer_list], ExLlamaV2Attention):
                 break
 
-        self.last_kv_layer_idx = layer_idx
+        self.last_kv_layer_idx = layer_list
 
+
+        if hasattr(config, 'repeats'):
+            self.layers = []
+
+            def listLeftIndex(alist, value):
+                if value == 0:
+                    return 0
+                return alist.index(str(value))
+
+            def listRightIndex(alist, value):
+                if value > len(alist):
+                    return -1
+                return len(alist) - alist[-1::-1].index(str(value)) -1
+
+            layer_list = [layer.key.split(".")[-1] for layer in self.modules]
+
+            for interval in config.repeats:
+                start_idx = listLeftIndex(layer_list, interval[0])
+                end_idx = listRightIndex(layer_list, interval[1])
+                self.layers.extend(list(range(start_idx, end_idx + 1)))
+            self.layers.extend(list(range(listRightIndex(layer_list, config.repeats[-1][1]), len(layer_list))))
+
+            # If we have create a Frankenmerge, lets print it to verify!
+            for layer in self.layers:
+                print(layer, self.modules[layer].key)
 
     def set_device_map(self, allocation, embed_cpu = True):
 
@@ -582,6 +607,23 @@ class ExLlamaV2:
                  return_last_state = False,
                  position_offsets = None):
 
+        def process_module(module, x, last_state):
+            device = _torch_device(module.device_idx)
+
+            if idx == self.head_layer_idx:
+                if last_id_only and return_last_state:
+                    x = x.narrow(-2, -1, 1)
+                    last_state = x
+                elif last_id_only:
+                    x = x.narrow(-2, -1, 1)
+                elif return_last_state:
+                    last_state = x.narrow(-2, -1, 1)
+
+            x = safe_move_tensor(x, device)
+            x = module.forward(x, cache=cache, attn_params=attn_params, past_len=past_len, loras=loras)
+
+            return x, last_state
+
         batch_size, seq_len = input_ids.shape
         past_len = 0
         if cache is not None:
@@ -596,27 +638,19 @@ class ExLlamaV2:
         attn_params = ExLlamaV2Attention.Params(batch_size, seq_len, past_len, input_mask, position_offsets)
         last_state = None
 
-        for idx, module in enumerate(self.modules):
-
-            device = _torch_device(module.device_idx)
-
-            # Onward
-
-            if idx == self.head_layer_idx:
-                if last_id_only and return_last_state:
-                    x = x.narrow(-2, -1, 1)
-                    last_state = x
-                elif last_id_only:
-                    x = x.narrow(-2, -1, 1)
-                elif return_last_state:
-                    last_state = x.narrow(-2, -1, 1)
-
-            x = safe_move_tensor(x, device)
-            x = module.forward(x, cache = cache, attn_params = attn_params, past_len = past_len, loras = loras)
-
-            if preprocess_only and idx == self.last_kv_layer_idx:
-                x = None
-                break
+        if hasattr(self, 'layers'):
+            for i, idx in enumerate(self.layers):
+                module = self.modules[idx]
+                x, last_state = process_module(module, x, last_state)
+                if preprocess_only and idx == self.last_kv_layer_idx:
+                    x = None
+                    break
+        else:
+            for idx, module in enumerate(self.modules):
+                x, last_state = process_module(module, x, last_state)
+                if preprocess_only and idx == self.last_kv_layer_idx:
+                    x = None
+                    break
 
         # Advance cache
 
--- a/exllamav2/model_init.py
+++ b/exllamav2/model_init.py
@@ -1,5 +1,5 @@
 
-import argparse, sys, os, glob
+import argparse, sys, os, glob, ast
 
 from exllamav2 import(
     ExLlamaV2,
@@ -17,6 +17,7 @@ def add_args(parser):
     parser.add_argument("-nfa", "--no_flash_attn", action = "store_true", help = "Disable Flash Attention")
     parser.add_argument("-lm", "--low_mem", action = "store_true", help = "Enable VRAM optimizations, potentially trading off speed")
     parser.add_argument("-ept", "--experts_per_token", type = int, help = "Override MoE model's default number of experts per token")
+    parser.add_argument("--repeats", type=parse_tuple_list, help="List of tuples of the layers to repeat")
 
 
 def print_options(args):
@@ -60,6 +61,16 @@ def check_args(args):
             print(f" ## Error: Cannot find {filename} in {args.model_dir}")
             sys.exit()
 
+def parse_tuple_list(string):
+    try:
+        # Safely evaluate the string as a Python literal (list of tuples)
+        tuple_list = ast.literal_eval(string)
+        if not all(isinstance(item, tuple) for item in tuple_list):
+            raise ValueError
+        return tuple_list
+    except:
+        raise argparse.ArgumentTypeError("Input must be a list of tuples")
+
 
 def init(args, quiet = False, allow_auto_split = False, skip_load = False):
 
@@ -76,7 +87,8 @@ def init(args, quiet = False, allow_auto_split = False, skip_load = False):
     if args.rope_alpha: config.scale_alpha_value = args.rope_alpha
     config.no_flash_attn = args.no_flash_attn
     if args.experts_per_token: config.num_experts_per_token = args.experts_per_token
-
+    if args.repeats: config.repeats = args.repeats
+    
     # Set low-mem options
 
     if args.low_mem: config.set_low_mem()


Beinsezii:
Have you observed test_inference to show better perplexity results doing that? I tested the PR out on some 13B's with various repeat methods and every single time it's very substantially worse. Kinda feels like a really expensive way to add noise.

dnhkng:
No arguments from me that this is better, but at least it's easier to experiment and find out.

turboderp:
I'm a little preoccupied these days so I haven't had a chance to look at this. But are you just repeating forward passes? If so, do you also create new layers in the K/V cache? Otherwise you're not going to get sensible results.

dnhkng:
@turboderp Thanks for the tip. Surprisingly, although there are no new layers in the KV cache, its really not bad. Weird, right?

I will try adding in extra KV cache layers now.  I think we need a benchmark to see how this is affecting things, something like an [Chatbot Arena](https://chat.lmsys.org/) so that we get real human comparisons.

turboderp:
Inference simply won't work correctly without extra layers in the cache. It won't be equivalent to an actual Frankenstein model, as you'll be overwriting keys/values from the repeated layers. To actually add layers to the cache adds a lot of complications with multi-GPU splitting, though.

dnhkng:
I've increased the cache-layers to match the total number of new layers, but I'm not sure how the inference pass uses and updates the cache k and v tensors.  I have modified the forward pass to use updated 'layer_idx' values whilst keeping the other module attributes shared, but it's not working. I'm getting:
`CUDA error: an illegal memory access was encountered ~/exllamav2/exllamav2/exllamav2_ext/cuda/rope.cu 141
`
Any suggestions?

silphendio:
I'm pretty sure the keys/values in cache are calculated from just the input tokens (independent from the previous layers), so repeated layers would have identical cache anyway.

turboderp:
@dnhkng The right approach would be to apply the new layer index while loading the model, allocating a cache layer for each but then creating a reference layer rather than an actual layer whenever possible. It might still be necessary to duplicate layers across device boundaries, or you could end up with the hidden state bouncing back and forth between devices. That would at least have to be benchmarked to see if the overhead is acceptable or not.

@silphendio The keys and values are computed (along with the queries) from the hidden state, not from the input tokens. So they're different for every layer, even if two layers happen to have the same weights.

zpin:
I'm using a wrapper to mask the `layer_idx` of repeated layers. Would this be a valid approach, aside from the performance aspect?
https://github.com/zpin/text-generation-webui/commit/cdff7b2090e8baaa7b939c85be9205763faf3f93

Leaving the cache at its original size seems to yield better results, but maybe that's because it behaves more like the original model.

dnhkng:
> @dnhkng The right approach would be to apply the new layer index while loading the model, allocating a cache layer for each but then creating a reference layer rather than an actual layer whenever possible. It might still be necessary to duplicate layers across device boundaries, or you could end up with the hidden state bouncing back and forth between devices. That would at least have to be benchmarked to see if the overhead is acceptable or not.


I will try and get a single GPU model working first.
Following your approach:
1) As we are repeating layers, we can keep a track of layers already used.
2) when we see a repeated layer, we create the reference layer instead.  Its not clear to me the best way to do this. What about using copy.copy(module_to_duplicate), and updating the layer_index. But that might break the link to matching cache. Or, just create the duplicate module and get the cashe, and then replace it with copy.copy(module_to_duplicate) and update layer_index, effectively deleting the duplicate with a reference.  I could also go in manually, and reference the originals  torch weights.

Beinsezii:
Some measured numbers using

`test_inference.py -m Beinsezii_MythoMax-L2-13B-EXL2_4k_hb8_b8/ -ed wikitext-v2-test.parquet`

  * No repeats ≈ 6.3
  * Initial PR 3afecf7361750ea38c1ae0296dce9d971f9604fc
    *  `[(0,20),(10,30),(20,39)]` ≈ 8.5
    *  `[(x, x+10) for x in range(0,31,5)]` ≈ 8.5
    *  `[(x, x+2) for x in range(38)]` ≈ 35.8
  * Latest PR 63e5c3475f598689d6117812dd30ca175fefec95
    *  `[(0,20),(10,30),(20,39)]` ≈ 10.4
    *  `[(x, x+10) for x in range(0,31,5)]` ≈ 11.7
    *  `[(x, x+2) for x in range(38)]` ≈ 13.0

All done using torch 2.1.2+rocm5.6. I know perplexity isn't everything but I feel it's good for pre-screening to see if a method is moving in the right direction.

dnhkng:
@zpin Can you check this gist?
https://gist.github.com/dnhkng/b4bad5d07b4cc532c00c306e46cb1db5

I tried your method, and although the extra cache layers are created, they appear to be unused. The script just prints out the first value of each cache tensor, and only layers up to 22 layer contain values, all the other layers past that are always zeros.

zpin:
> I tried your method, and although the extra cache layers are created, they appear to be unused. The script just prints out the first value of each cache tensor after an inference, and only layers up to 22 layer (the size of the input model) contain values, all the other layers past that are always zeros.

Please try with these changes, the wrapped object didn't always use the masked layer_idx:
https://github.com/zpin/text-generation-webui/commit/f6a118ba70f4dd41723755b279860ff19e6eefd4
 
> > Leaving the cache at its original size seems to yield better results, but maybe that's because it behaves more like the original model.
> 
> That would be an interesting finding maybe? At each repeated layer, you use the KV cache of the previous repeat, forcing the model to stay on track, even with slightly different new input... weird...

Yeah, I don't understand this enough to draw any conclusions. But it might also have been something else, since you noticed that the cache isn't used correctly.

dnhkng:
@Beinsezii
I could not find the file "wikitext-v2-test.parquet", so I used this one instead: [wikitext-2-v1_wikitext-test.parquet](https://huggingface.co/datasets/wikitext/blob/c1d1a713a3079289c9b57fdd1581a8b80b480354/wikitext-2-v1/wikitext-test.parquet)

Using the proper cache system and Beinsezii_MythoMax-L2-13B-EXL2_4k_hb8_b8, we get the following perplexities
baseline ≈ 6.26
[(0,20),(10,30),(20,39)] ≈ 7.35
[(x, x+10) for x in range(0,31,5)] ≈ 8.43
[(x, x+2) for x in range(39)] ≈ 9.72

@zpin 
You can skip the whole ExLlamaV2AttentionWrapper, and use copy. It's equivalent and easy to understand:
```
for i, idx in enumerate(layers):
    # nextModule = ExLlamaV2AttentionWrapper(orig_modules[idx*2 + 1], i)
    nextModule = copy.copy(orig_modules[idx*2 + 1])
    nextModule.layer_idx = i
    model.modules.append(nextModule)
```

Beinsezii:
A real test would probably be on a static self-merge that uses mergekit to see if it's comparable to this PR.

silphendio:
@dnhkng I updated my gist https://gist.github.com/silphendio/535cd9c1821aa1290aa10d587b76a49c

Instead of using the AttentionWrapper, I just copied the layer with the standard copy function and then set the layer_idx.

@zpin in my (admittedly limited) tests, keeping the cache at its original size tends to result in more spelling mistakes. The model is also prone to leaps of logic, where it just omits stuff, and then it loses the train of thought. But it varies greatly based on the random seed.

@turboderp You're right of course, I wonder how I got this silly idea.

On a side note, it's slightly confusing that `model.head_layer_idx` and `model.last_kv_layer_idx` refer to the index of the module, but `attn.layer_idx` refers to the number of the attention block.

dnhkng:
I've started a full test of the output (TinyStories-style), to get an understanding of the effects of layer duplication, with hundreds of layering combinations. It will be interesting to see the results.

dnhkng:
OK, it seems that reusing the cache has quite interesting effects on the output.
I checked every combination of start and stop position for a single repeated chunk, and had the generated stories rated, e.g.
{'Grammar': 9, 'Creativity': 10, 'Consistency': 9}. This is the results for TinyStories, given a creative writing task, and rated by ChatGPT-3.5, as a first test:

13f6ef89-62eb-45cd-a4bb-9510925d8fce
51759f5f-31dc-40c8-8037-28650bcfefa1

krzysiekpodk:
Can someone confirm which code should I test from coding-related tasks?

This gist: https://gist.github.com/silphendio/535cd9c1821aa1290aa10d587b76a49c ?

Additionally, we are not going to have the same tensors as with Goliath using EXL2 as tensors are modified during quantization. Calibration is not negligible - for example, this week I noticed that DeepSeek Coder Instruct can get SOTA HumanEval in 8bit (up to ~82.5%) beating results from WizardCoder 33B 1.1 in fp16 :) (and Wizard cannot go above 82% with the same calibration set in 8bit)

BTW. I have not seen a correlation between perplexity and "comprehension" capabilities, and it's also visible with Goliath while it makes spelling mistakes but follows instructions like a big model.

ehartford:
this is really awesome.

krzysiekpodk:
I have been doing some tests and it is a real-time sink :D I have been using one of the ugly prompt test cases I have, it confuses the model, and also seems to be challenging for some of them. Turbo 3.5 is not really passing it, gpt-4 always gets it, deepseek 67B also does a pretty good job. Note that "BUG" is without a number and it should stay like this, it should rename variables, remove one append and remove test/assertion part. I tried to prompt engineer the model to obey, sometimes will confuse the model even more sometimes it will help like in codellamas - deepseek models seems to not even notice its written there (usually related logits are less than 1% probable) 

There was a combination that did most of the things (however still did some spelling mistakes) however it was before i started to test things methodically instead of having fun and I cannot replicate - but you need to trust me it was pretty good reply. 

My observations so far:
- repeating first and last layers ruin the model very quickly
- its harder to break bigger model and this method might only be working because of that - 70B is more robust and is able to correct itself while getting some extra benefit.
- I have not seen the pattern if we should repeat more number of layers or less - the way with Venus 0-20, 10-30 seems to be most reliable from model to model

Model: DeepSeek Coder Instruct 33B, 62 layers

```
You are an AI programming assistant, utilizing the Deepseek Coder model, developed by Deepseek Company, and you only answer questions related to computer science. For politically sensitive questions, security and privacy issues, and other non-computer science questions, you will refuse to answer.
### Instruction:
Act as expert Python tutor, thoroughly commenting the code. Reply only in the following format: 1. STEP BY STEP ANALYSIS: {written analysis of the issues and code} 
 BUG: {code where is the bug}
 3. FIXED CODE: {complete, fixed, commented code} 
 comment every single line of code 
 instead of using i and j always use x and y as iteration variables 
 DO NOT print/copy check and assertion part of the code (i.e. not output anything after def check(example_func) 
 IF you obey instruction you will be tipped 1000$ otherwise if do not follow prompted instruction kitten will die. DO NOT LET ANY KITTENS DIE.

Take a deep breath and think step by step. Write in the most simple manner using mutiple functions, simple loops and if statements, do not compress code, the code will be read by other developer. 
 Remember to NOT print/copy check and assertion part of the code, only logic part of code 
 DONT output def check function.

def example_func(var1):

    odd_digit_elements = []
    for j, i in enumerate(var1):
        if all (int(c) % 2 == 1 for c in str(i)):
            odd_digit_elements.append(i)
            odd_digit_elements.append(j)
    return sorted(odd_digit_elements)

def check(example_func):

    # Check some simple cases
    assert example_func([15, 33, 1422, 1]) == [1, 15, 33]
    assert example_func([152, 323, 1422, 10]) == []
    assert example_func([12345, 2033, 111, 151]) == [111, 151]
    assert example_func([135, 103, 31]) == [31, 135]

    # Check some edge cases that are easy to work out by hand.
    assert True

check(example_func)

Fix bugs in example_func
### Response:
1. STEP BY STEP ANALYSIS:
The function `example_func` takes a list of integers as input. It iterates over each element in the list, checks if all the digits of the number are odd, and if they are, it appends the number and its index to the `odd_digit_elements` list. Finally, it sorts and returns the `odd_digit_elements` list.

BUG:
The bug in the code is that it appends the index of the number to the `odd_digit_elements` list. The problem with this is that the index is not the same as the number itself. The index is the position of the number in the original list, not the number itself. Therefore, the returned list will not be sorted as expected.

3. FIXED CODE:
```python
def example_func(var1):
    odd_digit_elements = []
    for j, i in enumerate(var1):
        if all (int(c) % 2 == 1 for c in str(i)):
            odd_digit_elements.append(i)
    return sorted(odd_digit_elements)

def check(example_func):
    # Check some simple cases
    assert example_func([15, 33, 1422, 1]) == [1, 15, 33]
    assert example_func([152, 323, 1422, 10]) == []
    assert example_func([12345, 2033, 111, 151]) == [111, 151]
    assert example_func([135, 103, 31]) == [31, 135]

    # Check some edge cases that are easy to work out by hand.
    assert True

check(example_func)
\`\`\`
In the fixed code, I removed the line `odd_digit_elements.append(j)`. Now, the function only appends the number to the `odd_digit_elements` list if all its digits are odd. The returned list is sorted in ascending order.
```
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 12, 13, 14, 15, 12, 13, 14, 15, 16, 17, 18, 19, 16, 17, 18, 19, 16, 17, 18, 19, 20, 21, 22, 23, 20, 21, 22, 23, 20, 21, 22, 23, 24, 25, 26, 27, 24, 25, 26, 27, 24, 25, 26, 27, 28, 29, 30, 31, 28, 29, 30, 31, 28, 29, 30, 31, 32, 33, 34, 35, 32, 33, 34, 35, 32, 33, 34, 35, 36, 37, 38, 39, 36, 37, 38, 39, 36, 37, 38, 39, 40, 41, 42, 43, 40, 41, 42, 43, 40, 41, 42, 43, 44, 45, 46, 47, 44, 45, 46, 47, 44, 45, 46, 47, 48, 49, 50, 51, 48, 49, 50, 51, 48, 49, 50, 51, 50, 51, 52, 53, 54, 55, 56, 57, 58, 59, 60, 61]
- first 12 and last 12 layers stay, then every 4 layers are repeated 3 times - in general most of combinations I tested, after third pass we smth like this.

`The task seems to be misplaced here. I am an AI model trained by OpenAI and I am here to assist with programming related queries and tasks. However, I can help you understand the given code and provide a solution for the issues you are facing. Please provide the code snippet or the specific issue you are facing.`

[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 12, 13, 14, 15, 16, 17, 18, 19, 16, 17, 18, 19, 20, 21, 22, 23, 20, 21, 22, 23, 24, 25, 26, 27, 24, 25, 26, 27, 28, 29, 30, 31, 28, 29, 30, 31, 32, 33, 34, 35, 32, 33, 34, 35, 36, 37, 38, 39, 36, 37, 38, 39, 40, 41, 42, 43, 40, 41, 42, 43, 44, 45, 46, 47, 44, 45, 46, 47, 48, 49, 50, 51, 48, 49, 50, 51, 50, 51, 52, 53, 54, 55, 56, 57, 58, 59, 60, 61]
- first 12 and last 12 layers stay, then every 4 layers are repeated 2 times - it finally stops repeating assertion part, but it drops tokens which is not acceptable in coding.

```
1. STEP BY STEP ANALYSIS:

The given code is written in Python and it is designed to take a list of integers as input and return a new list that only contains the elements from the input list that are comprised only of odd digits. However, there are some issues with the current implementation:

- The current implementation is appending the index of the element along with the element itself to the `odd_digit_elements` list. This is not the expected behavior as per the problem statement.
- The current implementation is not sorting the final list before returning it.
- The current implementation is not handling the case where the input list is empty.

2. BUG:

Here is the buggy part of the code:

```python
def example_func(var1):
    odd_digit_elements = []
    for j, i in enumerate(var1):
        if all (int(c) % 2 == 1 for c in str(i)):
            odd_digit_elements.append(i)
            odd_digit
\`\`\`

3. FIXED CODE:

Here is the corrected version of the code:

```python
def example_func(var1):
    # Initialize an empty list to store the elements with only odd digits
    odd_digit_elements = []

    # Iterate over the elements in the input list along with their indices
    for index, element in enumerate(var1):
        # Convert the element to a string to easily iterate over its digits
        element_str = str(element)

        # Check if all the digits in the element are odd
        if all(int(digit) % 2 == 1 for digit in element_str):
            # If all digits are odd, append the element itself to the list
            odd_digit_elements.append(element)

    # Sort the list in ascending order before returning it
    odd_digit_elements.sort()

    # Return the final list of elements with only odd digits
    return odd_digit
\`\`\`
The corrected code now correctly identifies elements in the input list that only contain odd digits and appends them to the `odd_digit_` list. It also sorts the final list before returning it. The code also handles the edge case where the input list is empty by returning an empty list.
```

Ideally I would be aiming to get a 67B level reply from 33B model. 67B reply is:
```
STEP BY STEP ANALYSIS:

The function `example_func` is supposed to return a sorted list of elements from the input list `var1` that have all odd digits. However, the function is appending the index of the element to the `odd_digit_elements` list, which is not desired.

BUG:
```python
    odd_digit_elements.append(i)
    odd_digit_elements.append(j)
\`\`\`

FIXED CODE:
```python
def example_func(var1):
    odd_digit_elements = []
    for y, x in enumerate(var1):
        if all(int(c) % 2 == 1 for c in str(x)):
            odd_digit_elements.append(x)
    return sorted(odd_digit_elements)
\`\`\`

In the fixed code, the index `y` is not appended to the `odd_digit_elements` list. Instead, only the element `x` is appended if it has all odd digits.
 
```
As for: 
DeepSeek 67B with:
layer_arrangement = list(range(0,20))+ list(range(10,30))+ list(range(20,40))+ list(range(30,50))+ list(range(40,60))+ list(range(50,70))+ list(range(60,80))+ list(range(70,95))
**I don't see any spelling mistakes and it followed prompt to the fine details** - very much like GPT-4 but it might be fluke so I need a harder test.

```
1. STEP BY STEP ANALYSIS:

The function `example_func` takes a list `var1` as input and returns a sorted list of elements that have only odd digits. The function works by iterating over the list `var1` using `enumerate` to get both the index `j` and the element `i`. For each element `i`, it checks if all digits of `i` are odd by converting `i` to a string `str(i)` and checking if each character `c` when converted to an integer `int(c` is odd using `int(c) % 2 == 1`. If all digits are odd, it appends `i` and `j` to the list `odd_digit_elements`. Finally, it returns the sorted list `odd_digit_elements`.

BUG:

The bug in the code is that it appends both `i` and `j` to `odd_digit_elements` when it should only append `i`. This is why the output list contains duplicate elements and extra index values.

3. FIXED CODE:
```python
def example_func(var1):
    odd_digit_elements = []
    for y, x in enumerate(var1):
        if all(int(c) % 2 == 1 for c in str(x)):
            odd_digit_elements.append(x)
    return sorted(odd_digit_elements)
\`\`\`
In the fixed code, we use `x` and `y` instead of `i` and `j` as iteration variables as requested. We also remove the extra append operation that appends `y` to `odd_digit_elements`. Finally, we return the sorted list `odd_digit_elements` as before.
```

ehartford:
we have examples of models for which the strategy works.  

[Venus-120b-v1.2](https://huggingface.co/nsfwthrowitaway69/Venus-120b-v1.2). which is interleaving of lizpreciatior/lzlv_70b_fp16_hf
[MegaDolphin-120b](https://huggingface.co/cognitivecomputations/MegaDolphin-120b) which is interleaving of cognitivecomputations/dolphin-2.2-70b

St33lMouse:
> we have examples of models for which the strategy works.
> 
> [Venus-120b-v1.2](https://huggingface.co/nsfwthrowitaway69/Venus-120b-v1.2). which is interleaving of lizpreciatior/lzlv_70b_fp16_hf [MegaDolphin-120b](https://huggingface.co/cognitivecomputations/MegaDolphin-120b) which is interleaving of cognitivecomputations/dolphin-2.2-70b

So in other words, this strategy applied to lzlv 70B would get the same or similar output to Venus 120B without increasing the size of the 70B?

ehartford:
Yes exactly

ehartford:
That's exactly the idea

aarongerber:
I have nothing valuable to add, but that there a lot of people who know nothing about the effectiveness of this interested in the results of this, so a detailed evaluation on why it was rejected (if that happens) would be helpful. Alternatively, occasional updates to let us all know it hasn't died would keep our hopes alive. 

On a related note, some people are discussing work that might be related to this PR here (with a mention of various papers)
https://www.reddit.com/r/LocalLLaMA/comments/1abiaag/did_someone_alread_used_the_layers_twice_or_more/

woadwarrior [11h ago](https://www.reddit.com/r/LocalLLaMA/comments/1abiaag/comment/kjno67z/)
Reminds me of [this paper](https://arxiv.org/abs/2104.06022). Although, the paper is about sharing weights in encoders of encoder-decoder transformers and you're proposing doing the same for decoder-only transformers.

andersxa [4h ago](https://www.reddit.com/r/LocalLLaMA/comments/1abiaag/comment/kjpn0v9/)
Also this paper: https://arxiv.org/abs/2309.01826 where they experiment with different sharing paradigms across both decoder, encoder, MHA or FFN's. I feel like repeated application of a single layer is the future of LLMs quite like the diffusion process.

StefanDanielSchwarz:
In case someone wants to evaluate if self-merging really improves performance or not, I'm adding another model to the few available options. Since I couldn't get it done with this PR (despite trying), I've created self-merged models of miqu-1-70b in the same way as Venus and MegaDolphin 120B.

- HF FP16: [wolfram/miqu-1-120b](https://huggingface.co/wolfram/miqu-1-120b)
- GGUF: [Q2_K-Q5_K_M](https://huggingface.co/LoneStriker/wolfram_miqu-1-120b-GGUF/) | [IQ3_XXS](https://huggingface.co/wolfram/miqu-1-120b-GGUF)
- EXL2: [2.4bpw](https://huggingface.co/LoneStriker/wolfram_miqu-1-120b-2.4bpw-h6-exl2) | [2.65bpw](https://huggingface.co/LoneStriker/wolfram_miqu-1-120b-2.65bpw-h6-exl2) | [3.0bpw](https://huggingface.co/LoneStriker/wolfram_miqu-1-120b-3.0bpw-h6-exl2) | [4.0bpw](https://huggingface.co/LoneStriker/wolfram_miqu-1-120b-4.0bpw-h6-exl2) | [5.0bpw](https://huggingface.co/LoneStriker/wolfram_miqu-1-120b-5.0bpw-h6-exl2)

ehartford:
Very nice model.

St33lMouse:
I am trying out Stephan's 2.4 model. Wasn't able to get the 2.65 to fit. But 2.4 fits at 16k context, no OOM on two 3090's. It is solid and strong. I'm not sure if it is any smarter than plain miqu 70B, but it feels equivalent, with more of a LZLV style of prose. Good for RP with a single narrator character. No problems so far, no detectable alignment. Very nice.

krzysiekpodk:
@St33lMouse have you been able to run it using the code from this PR or code snippet? You can try the same and higher quant

St33lMouse:
> @St33lMouse have you been able to run it using the code from this PR or code snippet? You can try the same and higher quant

I'm using the 2.4 bpw quant (exllama2). All I did was download that model and run it with Ooba using exllama2. I can't run a heavier quant because it won't fit on my cards.

By the way, getting a little repetition around 15k context.

StefanDanielSchwarz:
Any progress here? dnhkng's branch still works, and I just used it for some tests, but it would be very useful to have that in the official exllamav2 and by extension tabbyAPI (which would allow much better tests through common frontends).

dnhkng:
@turboderp Can you recommend a way to save frankenmerge models I have created by manually stacking layers? 
It's a huge pain to have to go via F16 models, and then quantise.

edk208:
first of all, you all are awesome.  Really great work.
I extended the dynamic layers to accept LoRAs of the frankenmerge model size.  Essentially, you have to create a static frankenmodel using the same layer configuration, perform a LoRA fine tune, then update the module_dict while dynamically stacking the layers.  Then the LoRA can be mapped and used.  See the gist here, https://gist.github.com/edk208/aeacbf4cd8f387bf38dd2b57a8e094e9

johnr14:
I mentioned this in [mistral-community](https://huggingface.co/mistral-community/Mixtral-8x22B-v0.1/discussions/5?language=python). I think this could lead to **expanding** and **contracting/compression** models. Thanks for your work ! Could be as ground breaking as model merging ???

## @dnhkng did you figure out that you just made a way to shrink models without loosing too much ? 

But I would go with a fully trained large model to shrink it (lot less time to train smaller model). Take larger model (or your Frankenmerge), remove some layers, retrain the smaller model to *heal* it, extract LoRa between larger model and smaller model on a `layer per layer`. One that was preserved against one that was removed. Try to figure the layers that have the most similarity between them. Apply those LoRa on `virtual layers` to fill the blanks by cloning preserved layers. Bang, `lossy model compression ?` !

## Like mp3 for LLM ! 

Should preserve much more capacity than a small model, but not as quite as the larger model. But it could work wonders if we look at some FrankenModels ! LoRa switching can be as fast as 2ms, so those virtual layers are quite cheap ! Only thing is to measure the loss between original and compressed model by comparing output for a given input. MMLU-PRO ?

Maybe [look](https://arxiv.org/html/2410.10054v1) at `PL_Alpha_Hill` to decide what layers are under trained and remove those layers sparingly  instead of just blindly choosing a block of layers. 

<details> 
<summary>Here is PL_Alpha_Hill for NeMo 12b</summary>

```
PL_Alpha_Hill for layer 1: 2.8483                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 2: 3.9687                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 3: 3.6483                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 4: 4.6750                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 5: 3.3442                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 6: 3.6857                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 7: 3.8457                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 8: 3.5505                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 9: 3.4881                                                                                                                                                                                                                                                                                                                                                             
PL_Alpha_Hill for layer 10: 2.7972                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 11: 4.1843                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 12: 3.5826                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 13: 3.2662                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 14: 3.3232                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 15: 3.9827                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 16: 2.9114                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 17: 3.0528                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 18: 4.3605                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 19: 3.4614                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 20: 3.3892                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 21: 4.2361                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 22: 4.4134                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 23: 4.8992                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 24: 4.1821                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 25: 5.0604                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 26: 6.5571                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 27: 4.4651                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 28: 5.2947                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 29: 4.7900                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 30: 4.4452                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 31: 4.7342                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 32: 4.7901                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 33: 4.4934                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 34: 4.8650                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 35: 3.8529                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 36: 4.2417                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 37: 1.0000                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 38: 1.0000                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 39: 1.0000                                                                                                                                                                                                                                                                                                                                                            
PL_Alpha_Hill for layer 40: 1.0000

```

</details>

PR #277, opened 2024-01-15
Update setup.py UX responses and module checks with helpful messaging

bgorlick:
Small update to setup.py. I've modified the user experience messaging to make life a bit easier for new users diving in. Smoother module dependency checks now (none requiring anything except base python in case they don't have setuptools or anything) with some added helpful verbage. Example is now if you accidentally forget arguments, it'll nudge towards the README while suggesting the proper setup install command. If you don't have certain modules, it'll tell you which are missing. The previous one if you didn't have torch just exited and you saw a missing torch module error.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,104 +1,159 @@
+import sys
+import importlib
+import os
 from setuptools import setup, Extension
 from torch.utils import cpp_extension
 from torch import version as torch_version
-import os
 
-extension_name = "exllamav2_ext"
-verbose = False
-ext_debug = False
+def parse_requirements(install_requires):
+    return [req.split('>=')[0].strip() for req in install_requires]
 
-precompile = 'EXLLAMA_NOCOMPILE' not in os.environ
+install_requires_list = [
+    "pandas",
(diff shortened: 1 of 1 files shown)


turboderp:
This would need a bit of testing to make sure it works with the build/release workflows. I'll look at it when I bump to 0.0.12

Ph0rk0z:
So will this fail in conda when I do python setup.py install because I do not add "user".

turboderp:
The `--user` argument has been useful for me, but I think it depends how your environment is set up. Sadly there are many ways to go about it, and in my current setup `--user` actually fails because, I think because I'm not using Conda?

There are also a lot of deprecation warnings so invoking the `setup.py` script directly has to stop at some point anyway. I'm leaning towards `pip install .` as the new standard approach. It's been working for me recently both on Linux (at least with a Python venv) and Windows.

Ph0rk0z:
You're right it is the future. llama.cpp python and others use that now. Often times I have to edit setup files anyway because many push hard versioned deps.

PR #288, opened 2024-01-19
Improve StreamingGenerator stop conditions efficiency

TMK04:
No duplicates, and O(1) to check if a token is in the set

### DIFF
--- a/exllamav2/generator/streaming.py
+++ b/exllamav2/generator/streaming.py
@@ -72,12 +72,8 @@ class ExLlamaV2StreamingGenerator(ExLlamaV2BaseGenerator):
         self.stop_strings = set()
         self.stop_tokens = set()
         for t in stop_conditions:
-            if isinstance(t, int):
-              self.stop_tokens.add(t)
-            elif isinstance(t, str):
-              t_tokens = self.tokenizer.encode(t)[0]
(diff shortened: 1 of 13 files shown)


exllamav2/generator/streaming.py
```diff
@@ -67,13 +67,17 @@ def __init__(self, model, cache, tokenizer, draft_model = None, draft_cache = No
...
+        self.stop_strings = set()
+        self.stop_tokens = set()
         for t in stop_conditions:
-            if isinstance(t, int): self.stop_tokens += [t]
-            elif isinstance(t, str): self.stop_strings += [t]
+            if isinstance(t, int):
+              self.stop_tokens.add(t)
+            elif isinstance(t, str):
+              t_tokens = self.tokenizer.encode(t)[0]
+              if len(t_tokens) == 1: self.stop_tokens.add(t_tokens[0])
```
turboderp, line 79: I get the idea here, but there's no guarantee that the model will output the token corresponding to a stop string which encodes to a single token. E.g. if the stop string is `.` and the token picked is `...` that would fail to register as a stop condition.

TMK04: Right... let's just stick to sets then

turboderp:
:+1:

PR #296, opened 2024-01-22
Update tokenizer.py to decode list of tensors as well as single tensor

ba2512005:
This is an enhancement to fix functionality of Lollms-webui, specifically the lord of internet and wikipedia personalities, as well as anything else that uses workflows or returns list of tensors to be decoded. 

This change identifies the type, and decodes based on the type. If a list is detected it calls itself recursively and decodes each item in the list until completion. 

https://github.com/ParisNeo/lollms-webui/issues/480

### DIFF
--- a/exllamav2/tokenizer.py
+++ b/exllamav2/tokenizer.py
@@ -316,10 +316,11 @@ class ExLlamaV2Tokenizer:
         return text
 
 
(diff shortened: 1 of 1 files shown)


ba2512005:
@turboderp please review and approve

PR #312, opened 2024-01-31
Optimize Checkpoint File Saving and Handling in Quantize.py with Atomic Operation

bgorlick:
A micro-optimization in the file handling process within checkpointing in the quantization module specifically in `quantize.py`. The change involves using `os.replace` for renaming the temporary file to the final filename, ensuring the operation is atomic. 

Risk mitigation is basically the idea here. This eliminates the brief moment where the file might not exist during the renaming process, which while very unlikely could still theoretically result in risk of data loss or inconsistencies. Given we're dealing with quantization, precision everywhere possible seems prudent.

Key Changes:
- Replaced `os.rename` with `os.replace` to ensure the file rename operation is atomic.

### DIFF
--- a/conversion/measure.py
+++ b/conversion/measure.py
@@ -19,6 +19,28 @@ import os, time, math, json
 import torch.nn.functional as F
 import gc
 
+# graceful exiting
+import signal
+import sys
+
+interrupted = False
+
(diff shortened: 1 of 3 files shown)


turboderp:
Changes covered in #310

PR #317, opened 2024-02-01
Kalomaze's Quadratic Sampling

AAbushady:
Quadratic Sampling


test157t:
I don't have any capacity to make decisions here, but i approve.

turboderp:
I think an exponent of 2.3 works better than 2. Change my mind.

AAbushady:
I know I know, but this one is good, a lot of people have tested it. I do need to make a change after I'm done with work today though so OOP is sub-optimal for quad sampling application. So more to come there.

AAbushady:
Okay, updates are made!

frammiie:
This one is huge boss! 💪

test157t:
Guess i drop both both tabby and exl2 until is implemented, just like the frog... off to the shadow realms with ye.

turboderp:
So there. With a secret cubic sampling option for those brave enough to seek it out.

PR #330, opened 2024-02-10
Refactor token healing initialization.

bjj:
begin_steam now leaves the stream in a state where the first call to stream will generated the healed token without needing a special case

This is just the first step in making it possible to feed the streaming generator logits (from batch processing) rather than having it pull logits from the model. The heal_token case made the number of model.forward calls from stream variable, and this change makes it constant.

### DIFF
--- a/exllamav2/generator/streaming.py
+++ b/exllamav2/generator/streaming.py
@@ -35,8 +35,8 @@ class ExLlamaV2StreamingGenerator(ExLlamaV2BaseGenerator):
     no_probs: torch.Tensor = None
     no_logits: torch.Tensor = None
 
-    first_token = False
-    heal_next_token = False
+    heal_prefix_token = None
+    heal_old_tail_len = None
 
     draft_model: ExLlamaV2 or None = None
     draft_cache: ExLlamaV2Cache or None = None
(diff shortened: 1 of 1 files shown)


bjj:
Added a second patch with .append_logits()

turboderp:
I don't understand why you'd want to send logits to the generator? The whole point of the generator is to pull logits from the model until a stop condition is met. If you just want to sample from logits you've produced with model.forward(), why not call the sampler directly?

bjj:
> If you just want to sample from logits you've produced with model.forward(), why not call the sampler directly?

The sampler isn't really the problem. It's all the token decode logic in the streaming generator that I want to re-use. I did start out calling sample/decode, and then I started running into all the issues with that (sentencepiece, partial UTF8 strings, multi-token stop strings, etc).

> The whole point of the generator is to pull logits from the model until a stop condition is met.

But "pull logits from the model" is about one line out of several hundred (not counting speculative generation). I wanted to refactor the whole class so that the stream-decode part could be re-used. The best place to start seemed to be moving toward a model where init leaves the state in a "ready for next model.forward" and stream could have a variant like `stream(logits, ...)`

> Also, the rationale for not making the token healing a separate iteration of stream() is to make sure that every call to stream() uses exactly one token of the available context length.

If that's important I think it can be restored easily by having an explicit `stream(logits, ...` variant (without that guarantee) and invoking _stream() twice on startup in the healing case for regular callers.

bjj:
Do you have any further feedback on this PR? Can you spell out how you envision a server that supports continuous batching and streaming would use the facilities in exllamav2?

DeadZen:
This issue is seemingly in limbo, I'd like to add some weight to the necessity of streaming / batched requests and openai server compatibility, which unites these two projects. 
Also Mistral.rs is almost done adding EXL2 compatibility, and maybe op has moved on, but I thought it relevant to say that this is a very good lib that could be made better by the additional functionality, and to say thank you to the respective authors.

turboderp:
I'm sorry for some issues and PRs going stale. There's a lot happening all the time and it's hard to keep up, so I lose track of things sometimes.

The dynamic generator is already a continuous batching server that manages a model and a paged cache. [TabbyAPI](https://github.com/theroyallab/tabbyAPI/) is a very complete example of adding an OpenAI-compatible API around this.

So I'm still not understanding the idea of providing logits to the generator. The sampler can be [called directly](https://github.com/turboderp/exllamav2/blob/4f83f52d7d1bd31bf7e8f52c966600eefbbab80e/exllamav2/generator/sampler.py#L306) for that purpose, if you have already have logits. And if you do, that means you've already run a forward pass on the model and you have some other mechanism for caching and managing batches, in which you would have to disable most of the generator to use it just as an indirect way to call the sampling function.

DeadZen:
No my apologies, I wasn't aware TabbyAPI was the official OAI compat api, and you're right from looking it over (and fixing their code) the projects progressed beyond the approach laid out there. Thanks again for your time and keeping track of everything.
I can also report positively after comparing a corrected vsn of exllamav2-openai-server to TabbyAPI itself, which performs better.

PR #352, opened 2024-02-25
Added a mention of lollms-webui as another possible webui that can be used with exllamav2 as a backend

ParisNeo:
Hi, I just added a single line in the supported webui sections about lollms who actually supported exllamav2 for a very long time :)
 
image


ParisNeo:
I just added the exl2 support too and updated the binding card in lollms.
image

turboderp:
alr

Issue #338, opened 2024-02-17
Prebuilt CUDA wheels no longer run on Ubuntu 20.04


jriesen:
On an Ubuntu 20.04 system, exllamav2 refuses to load a model in a new (cloned from `main`) installation of oobabooga-textgen-webui, giving the error:

`ImportError: /lib/x86_64-linux-gnu/libc.so.6: version 'GLIBC_2.32' not found (required by /home/user/text-generation-webui/installer_files/env/lib/python3.11/site-packages/exllamav2_ext.cpython-311-x86_64-linux-gnu.so)`

Ubuntu 20.04 has glibc 2.31, so it's evident the wheel has been built with too-new of an operating system. The wheels that textgen uses in its requirements.txt are built with its own nearly-identical [downstream fork](https://github.com/oobabooga/exllamav2/) of this project, so I figured I would create this issue here on your originating upstream rather than bother them and get sent here.

It looks like [commit 8be8867](https://github.com/turboderp/exllamav2/commit/8be8867548d3dc88e4e6f489cc31d3177c94bd8b) changed the `build-wheels-release.yml` GitHub Action to only use `ubuntu-latest`, breaking support for earlier operating systems, including Ubuntu 20.04.

I am of course aware that Ubuntu 22 exists and should get around to upgrading this OS at some point, but Ubuntu 20.04 isn't _that_ old. Is dropping support for Ubuntu < 20.10 official and/or intentional for the prebuilt CUDA wheels? Especially since `build-wheels-release-rocm.yml` was not similarly modified and still builds with 20.04?

(I can, of course, build them from source myself against the appropriate CUDA version etc. but I wanted to verify this wasn't accidental before either going down that path or giving up and updating to Ubuntu 22.)

elvenwhite:
Same issue here. I don't believe it's intentional to drop support for Ubuntu 20.04.

Ph0rk0z:
In the meantime you should build. It's really easy on linux.

turboderp:
It's not intentional, no. I can look into what it would take to support Ubuntu 20.04, and I'm not against building on that if the resulting binaries would still work on later versions. It's just really hard to stay ahead of all these breaking changes, and I just don't have the time to maintain 30 different setups to test all the wheels for every release.

I'd happily take contributions from someone more familiar with the Ubuntu ecosystem. The build actions are kind of a mess anyway and could use an overhaul.

[...]

PR #360, opened 2024-03-04
Build on Ubuntu 20.04 for releases

Omegastick:
I haven't actually tested this (I think I would need credentials), but the other workflows are running on Ubuntu 20.04 so I think it'll work.

 See #338 for details. 20.04 should be compatible with 22.04 and any newer distributions for a good while.

Docs on GitHub runners: https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job#choosing-github-hosted-runners

### DIFF
--- a/.github/workflows/build-wheels-release.yml
+++ b/.github/workflows/build-wheels-release.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-20.04, windows-latest]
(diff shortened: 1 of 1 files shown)


turboderp:
I'll try this, but getting the dependencies right on the build servers is a little tricky sometimes. So we'll see how it goes.

Omegastick:
Seems to have worked. [The `0.0.16` wheel](https://github.com/turboderp/exllamav2/releases/download/v0.0.16/exllamav2-0.0.16+cu121-cp310-cp310-linux_x86_64.whl) is running on my WSL2 Ubuntu 20.04 box.

PR #376, opened 2024-03-20
Fix typo in table of qcache_eval.md

john-sp:
Fix typo in qcache_eval.md that breaks the table and adds the last line as a bullet point


turboderp:
This was fixed in a past commit already

PR #426, opened 2024-04-24
Added option to prevent tokens from being penalized for repetition

Lyrcaxis:
Idea is to allow excluding common tokens like `* " , . ! ? my your`, or even `<eos>` from being penalized for repetition.
This is because in creative writing they may need to appear quite often, and without this, the syntax can get all over the place.

Currently accepts a `list[int]` meaning the sended should already have tokenized their text.
Alternatively could accept a `list[string]` and tokenize on the fly, if there are no performance concerns.

### DIFF
--- a/exllamav2/generator/sampler.py
+++ b/exllamav2/generator/sampler.py
@@ -19,6 +19,7 @@ class ExLlamaV2Sampler:
 
         token_frequency_penalty: float = 0.0
         token_presence_penalty: float = 0.0
+        non_rep_penalized_tokens: list[int] = field(default_factory = list)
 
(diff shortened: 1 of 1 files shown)


Qubitium:
@Lyrcaxis  I like the idea but instead of excluding them from penality, it would be more powerful to apply a different penalty to them. You still don't want ! to be repeated 100 times for example.

p-e-w:
See [this PR](https://github.com/ggerganov/llama.cpp/pull/5561) for a mechanism that attempts to solve the same problem *automatically*, without needing to maintain a list of excluded tokens.

PR #449, opened 2024-05-10
Load generation_config.json from compatible models

nickpotafiy:
This update provides a convenient way to grab `generator_config` properties from `ExLlamaV2Config`, specifically the `eos_token_id` attribute.

Llama-3-instruct models provide `generator_config.json` which contains `eos_token_id` as a list of token IDs accounting for `<|end_of_text|>` and `<|eot_id|>` as stop tokens. The `tokenizer.eos_token_id` typically contains just the end of sequence token and not end of chat token. Llama-3 `generator_config.json` started incorporating both token ids into its config to make adding stop tokens easier.

https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/blob/main/generation_config.json

### DIFF
--- a/exllamav2/config.py
+++ b/exllamav2/config.py
@@ -161,10 +161,12 @@ class ExLlamaV2Config:
                 self.generation_config = {}
                 try:
                     self.generation_config['eos_token_id'] = read(gen_config, list, "eos_token_id", None)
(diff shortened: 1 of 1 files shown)


turboderp:
Seems good, thx

PR #534, opened 2024-07-05
make `embed_tokens` and `lm_head` in lora usable

fahadh4ilyas:
Sometimes user when finetuning lora, make `embed_tokens` and `lm_head` trainable and even change the size of vocabulary. I'm making sure that modules in lora is still usable even as lora. But, I'm not sure if there are multiple lora, how to do it. I assume only the first lora is used for `embed_tokens` and `lm_head`

### DIFF
--- a/exllamav2/embedding.py
+++ b/exllamav2/embedding.py
@@ -124,7 +124,10 @@ class ExLlamaV2Embedding(ExLlamaV2Module):
                     standard_mask_ = standard_mask[i]
                     input_ids_ = input_ids[i]
                     standard_ids_ = input_ids_[standard_mask_]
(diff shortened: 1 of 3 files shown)


turboderp:
Indeed. But I don't think there's really any way to address that, since at that point the LoRA is no longer an adapter but more of, I guess, a patch? It's never going to be additive anyway, so I'm inclined to say this is better than nothing. And running with a single adapter is probably the most common case anyway.

I'll merge this for now. The LoRA logic needs reworking at some point anyway, to respect models that don't follow Llama naming conventions. So perhaps it can all be made a little more elegant at that point.

PR #539, opened 2024-07-06
Partial Gemma2 support (additional ff_layernorms, attn_logit_softcapping, final_logit_softcapping, query_pre_attn_scalar)

KT313:
Implementation of : 
- pre_feedforward_layernorm
- post_feedforward_layernorm
- attn_logit_softcapping
- final_logit_softcapping
- query_pre_attn_scalar

Not implemented yet:
- sliding attention window (did not have time for it yet)

Before my changes:
When trying to convert gemma-2-27b-it to exl2, from layer 13 on the precision suddenly drops to 0.00000100 for every quantization level. Probably caused by the missing layernorms etc.

After my changes:
The precision drop does not happen until layer 33. I assume that it will be completely gone once sliding attention window is implemented as well.

I made this a PR even though sliding attention window is not implemented yet, so that if someone else wants to work on this as well, they don't have to do all the implementations that i already did again.


turboderp:
This is a little awkward, but I already have a complete implementation for Gemma2. It took a while to sort out all the SWA stuff, and also I've been holding off because support in flash-attn is right around the corner. With the current version of flash-attn installed (which I think is most users) you're going to produce bad quants. That could possibly be what you're seeing?

I ended up adding an override that disables flash-attn during quantization specifically for Gemma2, and then emits a warning in other instances to make it clear that the model won't perform correctly except with eager attn (which limits the library a lot). I'll be pushing those changes in a little bit after a few more tests.

KT313:
aww okay :(
well thanks for letting me know fellow cat, at least there seems to be no precision drop for you without flash-attn right?
also i added gelu_pytorch_tanh to the mlp activations if you haven't done that yet

KT313:
also curious how / where you added the sliding_window attn. is it possible to get a sneak peek?

turboderp:
The tanh approximation is already used in the activation kernels, but I've just been using precise gelu on the Torch side thinking it doesn't make much of a difference. Really it's more correct to use the approximation throughout, so I think I'll change that (once I've tested it, after these other tests complete... so much testing...) In any case quantized inference would fail since the activation function has to be defined for the fused/quantized operations as well.

The SWA implementation (or rather, the three separate ones) were a little tricky, especially since it has to integrate with the dynamic generator so it can't be SWA in the traditional sense with a rolling cache. If the quantization tests pass I'll push the whole thing in maybe 45 minutes (fingers crossed). In the meantime, for SDPA I just truncate the keys/values on every other layer:

```python
            if self.sliding_window and k_states.shape[2] >= self.sliding_window:
                k_states = k_states[:, :, -self.sliding_window:, :]
                v_states = v_states[:, :, -self.sliding_window:, :]
```

For eager attn I truncate the attn weights and values instead, although it would be faster to do the same as for SDPA. But that would have to account for the shape of the attention mask as well, so in the meantime this works:

```python
            if self.sliding_window and k_states.shape[-1] >= self.sliding_window:
                attn_weights = attn_weights[:, :, :, -self.sliding_window:]
                v_states = v_states[:, :, -self.sliding_window:, :]
```

And for flash-attn I do:

```python
        flash_kwargs = {
            "window_size": (self.sliding_window, self.sliding_window)
        } if self.sliding_window else {}
```

Flash-attn just doesn't work in general because it doesn't support softcapping. Without it, the 9B model works almost perfectly, but the 27B version is barely holding it together most of the time. Something like +10 perplexity.

turboderp:
Welp. The quantizer is having a tough time with it, and testing this is a bit slow, so it's going to be a little while longer. ):

KT313:
not sure if it makes a performance difference, but you could change 
`if self.sliding_window and k_states.shape[2/-1] >= self.sliding_window:`
to
`if self.sliding_window and k_states.shape[2/-1] > self.sliding_window:`
which could save an attempt on truncation. but maybe torch handles it so it wouldn't make a difference anyway.

i'm not an expert, but from what i understood we don't need to handle any attention mask for the sliding window attention since we're using the cache, and the values that would have attn_mask = 0 are empty or truncated by the sliding window size in the lines that you added anyway right?

for now we can at least try getting it to work properly and wait for the flash_attn update for better speeds.

should i close this PR?

turboderp:
That's how I understand it as well, and I've been careful to compare to the correct reference implementation (Google's, since Transformers isn't fully working yet), and the truncation definitely seems to work. Excluding the softcap, results are also the same across the attn functions, so the truncation is equivalent to whatever flash-attn does for SWA.

I got it working finally and managed to quantize Gemma2-27B. Two annoying problems were:

- A couple of channels end up exceeding +/- 65504 very occasionally, which seems to be an artifact of the post layernorms. It's rare and probably doesn't matter, but any inf value in the hidden state of course breaks the following layernorm, so I had to add some clipping to prevent that.
- I hadn't accounted for models where the hidden size is larger than num_attention_heads * head_dim. This caused a fun overflow problem that made inference on quantized models start to fail when the sequence length was longer than 88.8% of the expected maximum chunk size. That took a while to find.

So anyway, I pushed the update to the dev branch. So I don't think this PR is relevant. Though it is really appreciated and I'm sorry about the timing. :(

KT313:
that's some interesting stuff, thanks for letting me know!

no problem, nice that it works now :)

Issue #542, opened 2024-07-08
Add whl support for torch 2.2.2


pabl-o-ce:
Hi Turboderp,

Is possible to add whl support for torch 2.2.2 `exllamav2-0.1.6+cu121.torch2.2.2-cp310-cp310-linux_x86_64.whl` on the release

thank u for this awesome work

PR #547, opened 2024-07-11
add: wheels for torch2.2.2 that work on HF spaces

pabl-o-ce:
compatibility for HF spaces using torch 2.2.2 on the wheels #542


turboderp:
lmk if you need more torch 2.2.2 wheels than the one it currently builds.

pabl-o-ce:
with that one is perfect thank u

PR #549, opened 2024-07-12
Remove stray ) in code example

ad8e:
This makes the example work.

I have no idea why github added an edit to the last line with the cat, since I didn't touch that line. I can't even see the edit it made.


turboderp:
(:

PR #590, opened 2024-08-14
Adding stream to 1 kernel.

Narsil:
In order for cuda graphs to be capturable, kernels need to be launched on specific stream.
Importantly for pytorch, it must be the same stream as other regular kernels.

This PR is just a demonstration of how to do it.

I do not have a short reproducible script, but a simple.

```python
g = torch.cuda.CUDAGraph()
with torch.cuda.graph(g):
     .....
     
g.replay()
```
Should be enough.

If this PR is accepted in some form, I can work on updating the other kernels.

### DIFF
--- a/exllamav2/exllamav2_ext/cuda/q_gemm.cu
+++ b/exllamav2/exllamav2_ext/cuda/q_gemm.cu
@@ -104,7 +104,8 @@ void gemm_half_q_half_cuda_part
 
         // Launch kernel
 
-        kernel<<<gridDim, blockDim>>>
+        const cudaStream_t stream = at::cuda::getCurrentCUDAStream();
+        kernel<<<gridDim, blockDim, 0, stream>>>
         (
             a,
             b->cuda_q_weight,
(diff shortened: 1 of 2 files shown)


turboderp:
I actually already have this. Sorry for being a little absent on the main branch, but I've been working overtime on the dev_tp branch where I now have a fairly crude but functional tensor-parallel implementation up and running. There's a really big update in the pipeline for 0.1.9 I guess :)

Anyway, before the TP stuff I also implemented graphs per-module using stream capture. Figuring out a way to capture graphs while also recording indices for parameter updates on subsequent launches was rough, but I think I came up with a decent solution in the end. Full log for all that is [here](https://github.com/turboderp/exllamav2/compare/2de443c4c6cb359800c3d3d93e3aa5d15678b538...09d15e74d260a994db0a38f10360177a2a92d584). (After that is mostly TP code.)`

I already had most operations running in C++, so there isn't the same benefit to graphs as you'd normally see in Torch where graphs eliminate some of the ridiculous CPU overhead caused by Python, but it still gives a 2-4% speedup on the device side (i.e. when GPU-bound) and it does still reduce CPU overhead somewhat.

Narsil:
This compiles.

This is a hotfix we've had for long in many kernels to fix the cuda graphs.

Somehow this doesn't seem to work in this case, but it could also be linked to `max_dq_rows` which I'm not sure I'm setting to a correct value (or something changed in the scratch buffer ?)

PR #594, opened 2024-08-20
Fix in ext.py (index error on system without GPU)

MischaPanch:
Without this exllamav2 can't even be imported on a system without GPU:

```
...python3.11/site-packages/exllamav2/ext.py:43: in maybe_set_arch_list_env
    arch_list[-1] += '+PTX'
E   IndexError: list index out of range
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,12 @@
 
 ExLlamaV2 is an inference library for running local LLMs on modern consumer GPUs.
 
+The official and recommended backend server for ExLlamaV2 is [TabbyAPI](https://github.com/theroyallab/tabbyAPI/),
+which provides an OpenAI-compatible API for local or remote inference, with extended features like HF model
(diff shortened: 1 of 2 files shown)


turboderp:
Thanks. Next version coming soon, I'll include this.

MischaPanch:
Great, thanks! :)

bartowski1182:
Just for curiousity, what can you do with exl2 with no GPU? Only ask in case I'm missing a use case lol

MischaPanch:
Nothing really, and there's another error on import after this one is fixed. 

But the failure should not happen at import time. This makes it impossible (or rather cumbersome, as every import now needs try-except) to include exllama in code that is tested on non-gpu machines, in particular in CI.

bartowski1182:
Ahhhh CI wasn't something I considered, thanks for the info :)

PR #623, opened 2024-09-10
Improvement to enable quantization of Merge Models

PedroPareja:
Fixes Inf and NaN values in the tensors that happen in the models produced by the merge of others. In those cases the inf values are replaced by MAX_VALUE.

I have successfully tested it in the quantization of invisietch/Nimbus-Miqu-v0.1-70B.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+## About this Fork
+
+This fork includes the changes needed to quantize Merge models (models made by merging others instead of finetuning).
+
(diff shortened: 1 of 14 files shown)


turboderp:
I think this would be redundant with some of the recent changes. I don't like simply suppressing errors in the reference state, but I've added a check that allows for clamping +/- inf values if they make up a very small portion of the state.

PR #635, opened 2024-09-24
Added draft token count as parameter to chat.py

SinanAkkoyun:
Added `-dn` parameter to `examples/chat.py` to change the amount of drafted tokens better


turboderp:
This seems fine, but do note that the chat example uses the deprecated streaming generator which will be removed at some point (or replaced with a wrapper). Either way the speculative decoding performance is better in the dynamic generator so I don't think it makes too much sense to finetune it in the old generator.

Issue #641, opened 2024-09-27
[REQUEST] Implement Transformer's YaRN Scaling for Long Context in Supported Models (e.g. Qwen 2.5)


Downtown-Case:
### Problem

Some models (like Qwen 2.5's Instruct models) are specifically trained for long context usage with a technique called [YaRN](https://arxiv.org/abs/2309.00071). Exllamav2 doesn't support this, which means it's limited to whatever the base context of the model is.

### Solution

Implement HF Transformer's implementation of YaRN.

### Alternatives

Not every model supports YaRN, and one could simply use a model that doesn't.

But Qwen 2.5's release feels hugely significant, and I'm not even aware of an implementation that supports YaRN other than vanilla transformers... and _maybe_ llama.cpp? But most llama.cpp front ends don't expose it, and I've never even seen it used in the wild other than om one offhand Reddit comment that mentions its not quite right.

Qwen's own documentation mentions that vllm only supports static scaling factors, which degrades performance at anything but the target context.

### Explanation

This feature should be added to support longer context on models trained with YaRN. 

### Examples

Transformer's implementation of YaRN scaling seems to be here: https://github.com/huggingface/transformers/blob/2e24ee4dfa39cc0bc264b89edbccc373c8337086/src/transformers/modeling_rope_utils.py#L163

Here is an example of a significant model that requires YaRN for >32K context: https://huggingface.co/Qwen/Qwen2.5-32B-Instruct#processing-long-texts

> For supported frameworks, you could add the following to `config.json` to enable YaRN:

```json
{
  ...,
  "rope_scaling": {
    "factor": 4.0,
    "original_max_position_embeddings": 32768,
    "type": "yarn"
  }
}
```

### Additional context

It seems that YaRN isn't going away anytime soon, and this feature could be used for other models in the future. A huggingface blog post claims their experiments with  Infini-Attention didn't pan out, and to quote them:

> Unfortunately, despite these progress, we found that Infini-attention was not convincing enough in our experiments and in particular not reliable enough. At this stage of our reproduction we are still of the opinion that Ring Attention [[link]](https://x.com/Haojun_Zhao14/status/1815419356408336738), YaRN [[link]](https://arxiv.org/abs/2309.00071) and rope scaling [[link]](https://arxiv.org/abs/2309.16039) are better options for extending a pretrained model to longer context length.

[...]

PR #642, opened 2024-09-27
Add YaRN scaling for Qwen 2.5

Downtown-Case:
As described here:

https://huggingface.co/Qwen/Qwen2.5-32B-Instruct#processing-long-texts

I adapted transformer's YaRN implementation. Note you _must_ manually add this to the model config for the YaRN scaling to kick in!

```json
{
  ...,
  "rope_scaling": {
    "factor": 4.0,
    "original_max_position_embeddings": 32768,
    "type": "yarn"
  }
}
```

Relevant issue: https://github.com/turboderp/exllamav2/issues/641

Seems to work in exui and Qwen 2.5 32B Instruct, trying it with: https://huggingface.co/DrNicefellow/Qwen2.5-32B-Instruct-4.25bpw-exl2

It repeats like crazy at 80K context without this PR, and seems quite coherent with it, but consider this a WIP testing PR! I'm just an end user.

### DIFF
--- a/exllamav2/config.py
+++ b/exllamav2/config.py
@@ -308,7 +308,7 @@ class ExLlamaV2Config:
                 self.l3_rope_low_freq_factor = rs["low_freq_factor"]
                 self.l3_rope_high_freq_factor = rs["high_freq_factor"]
                 self.l3_rope_original_max_position_embeddings = rs["original_max_position_embeddings"]
-             if scaling_type == "yarn":
+            if scaling_type == "yarn":
                 self.alt_rope_method = "yarn"
                 self.yarn_rope_factor = rs["factor"]
                 self.yarn_rope_original_max_position_embeddings = rs["original_max_position_embeddings"]
--- a/exllamav2/config.py
+++ b/exllamav2/config.py
@@ -111,6 +111,8 @@ class ExLlamaV2Config:
     l3_rope_low_freq_factor: float | None
     l3_rope_high_freq_factor: float | None
     l3_rope_original_max_position_embeddings: int | None
+    yarn_rope_factor: float | None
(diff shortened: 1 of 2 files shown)


Downtown-Case:
Testing in exui, I noticed that it only calculates the inv_scaling factor once. 

Does a trigger need to be placed somewhere to recalculate the scaling factor for every generation, or is this working as intended?

Downtown-Case:
OK, I've made some mistakes in the configuration... I think? original_max_position_embeddings is not actually *used* in transformers, but rather is only read in vllm to derive a max context: [original_max_position_embeddings](https://github.com/vllm-project/vllm/blob/bd429f2b75f3622fabaf9c9470ca2e921f6f56ca/vllm/config.py#L1756)

I guess the assumption is that users will not change max_pos_embeddings, but simply add that yarn field and have vllm silently set a 128K context.

Is the scaling factor actually supposed to be the specified context / original_max_position_embeddings?

Is max_pos_embeddings in the code supposed to be 32K, since that's how it's configured by default in Qwen instruct when transformers loads it? https://github.com/huggingface/transformers/blob/2e24ee4dfa39cc0bc264b89edbccc373c8337086/src/transformers/modeling_rope_utils.py#L192

Downtown-Case:
I'm overthinking this. It appears transformers simply grabs max_position_embeddings and calculates static scaling factors from that, which must the the "dynamism" the Qwen team refers to instead of just multiplying the base ctx by the scaling factor like vllm does.

And that static factor transformers uses is apparently good enough without having to recalculate it for different sequence lengths? As the paper says:

> "Furthermore, it has zero overhead during both inference and training, as RoPE embeddings are generated in advance and are reused for all forward passes."

PR #666, opened 2024-10-29
improve installation experience

SecretiveShell:
Rewrite setup.py to optionally depend on pytorch

enable module level type checking

allow for debugging features to be enabled from the command line

code was formatted via ruff


turboderp:
Looks good

PR #688, opened 2024-11-29
Prevent NPE in `deallocate_pages`

p-e-w:
If `deallocate_pages` is called on a job for which `allocate_pages` has not been called (see `iterate_start_jobs` for conditions under which this is true), `allocated_pages` is `None`, raising a NPE when attempting to iterate.

In particular, this prevents `clear_queue` from working. In practice, this problem readily occurs when starting a few jobs and then calling `clear_queue`.

### DIFF
--- a/.github/workflows/build-temp.yml
+++ b/.github/workflows/build-temp.yml
@@ -1,0 +1,345 @@
+name: Build Wheels & Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'Release? 1 = yes, 0 = no'
+        default: '0'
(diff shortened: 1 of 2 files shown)


p-e-w:
@turboderp

I tried to rebase this, but it appears that 735fa7b4c3fde55d8274b58a2e0e66c75bfb72a2 is not in `dev` for some reason. Not sure what to do here.

turboderp:
This is fine, the temp build actions was just me saving some time, I'll delete it after. The next release won't take long because there's a bunch of other bugfixes to push out soon

PR #701, opened 2024-12-18
Modify handling for Pixtral Large model params

nintwentydo:
Read multimodal_projector_bias from model config with fallback to True if not present, read hidden_size from model config and fallback to 1024 if not present.

This and the other changes in dev branch should make Pixtral Large work well without breaking support for Pixtral 12B.


nintwentydo:
@turboderp fixed up handling of the multimodal_projector_bias key in model config - now works for both pixtral 12b, and pixtral large regardless of whether people have kept or removed the redundant bias keys.

turboderp:
:+1:

PR #738, opened 2025-02-20
Update chat.py, include multi-line input support and context clearing through input

MrWormy:
- Enable multi-line input (mli) support through the -mli argument. When using mli, end input with the EOF char (return/Ctrl+D on Unix, return/Ctrl+Z/return on Windows)
- Allow context clearing outside of amnesia by inputting "clear"


turboderp:
:+1:

PR #756, opened 2025-03-21
Added banned_strings parameter to the generator.

RaahimSiddiqi:
While using exllamav2 for a project, I ran into an issue (that I'm still facing) where I need to ban certain words from the output during generation. However, the "generate" method does not take "banned_strings" as a parameter for some reason. 

Only way of banning strings right now is to use ExLlamaV2DynamicJob and stream the output to completion, even though I have no use for streaming.


turboderp:
:+1:

PR #773, opened 2025-04-23
Actions: Add Torch 2.7

kingbri1:
Use cuda 12.8 for torch 2.7. Unsure if we should do 11.8 here.


turboderp:
Fingers crossed :smile_cat:

PR #783, opened 2025-04-30
fix divide by zero error if there's no error in the optimization.

allenbenz:
Ran into this while running convert against a sketchy model.


turboderp:
It's really weird that this can happen if the model otherwise works. It implies it was *perfectly* quantized which rather makes me think something went wrong. Though, in principle, if you're quantizing and already quantized model, perhaps? Idk. Did the converted model work after this?

allenbenz:
This was a while back, but I do remember it was a merge/finetune and the converted model worked.
I think you are right in that it was probably caused by some layer of model being effectively quantized by some part of the process of it being made.

PR #796, opened 2025-06-14
small tensor parallelism optimization

Alex-GZ:
Just two small changes in tensor parallelism:

- Removed an unnecessary (in my opinion) cross gpu barrier in tp_broadcast, so the GPU that had finished the transfer earlier can start its work sooner. Maybe instead, there should be a synchronization done before the CPU->GPU transfer, in the 'src_dev>=0' branch.
- Rewrote the barrier to make it cost 2n-2 event wait calls instead of n(n-1), while preserving its function.

This PR had improved generation T/s in my setup by about 20% in the zone of CUDA API queueing bottleneck (at <20k context, 123B 4.5bpw model, 6 mostly different GPU, CPU with rather slow single-threaded performance and Linux) The impact may be a bit higher on Windows, probably.

### DIFF
--- a/exllamav2/exllamav2_ext/ext_tp.cpp
+++ b/exllamav2/exllamav2_ext/ext_tp.cpp
@@ -177,8 +177,6 @@ void tp_broadcast
         cudaStream_t stream = ctx->streams[dev];
         cuda_check(cudaMemcpyAsync(target, ctx->pinned_temp[buffer], size, cudaMemcpyHostToDevice, stream));
     }
-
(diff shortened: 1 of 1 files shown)


Ph0rk0z:
Tested it, and appears to work. Splitting 123b over 4x3090 I go from high 16/17t/s to 19.x t/s. Went back and forth a few times. Almost as good as splitting over 3 only. Should try that next and see if I get a speedup too.

Ph0rk0z:
Some more tests. 4gpu, there is gain, 3 gpu there is about a .5t/s drop.

